Issue #47, opened 2017-07-04
npm is not installed


tnguyen14:
Hi,

I just downloaded Terminus for the first time tonight on my Windows machine (with WSL), and I'm really impressed. It got a lot of nice things right.

I was also impressed with the long list of themes available right out of the box.  Well done!

I have a few simple questions that I was hoping you could help with:
- Can a new theme be added? I was pleasantly surprised when I saw the Tomorrow Night themes. Those have now been superseded by Base16 themes (see https://github.com/chriskempson/base16), which is a very popular project. Could support for that be added?
- On the plugins page, it says that NPM is not installed, even though I do have npm installed on WSL. But it's probably not where terminus is looking for it. Where should I have node and npm installed for it to pick up?
- Is there a way I could set the default path for a new terminal WSL windows? Right now, it is set to ` /mnt/c/Users/tri/AppData/Local/terminus/app-1.0.0-alpha15`. I wonder if that could be changed to home (`~`).

Thank you very much for the amazing work put into this so far. Please let me know if there's anything I could help with.

tnguyen14:
I'm also aware that these are 3 questions combined into a single issue. I thought I'd do that to avoid spamming the repo. If you'd prefer they be split into their own issues, I'm more than happy to do that.

Eugeny:
1. Currently it has to be either built as a plugin, or installed as a Hyper theme. I'm working to add an ability to automatically pick up themes from other sources on the system
2. `npm` should be available on `$PATH` for the detection to work. This works with e.g. the official NodeJS installer (linked directly inside Terminus).
3. Please open this as a separate bug and I'll take a look!

tnguyen14:
I've opened #60 for 3.

For 2, I have node and npm available on `$PATH` inside WSL. It could be that Terminus looks for it in Windows system path somewhere? I'll try installing Node directly on Windows and report back.

Eugeny:
Yes, it needs the Node to be installed in Windows itself, because it will then use npm to install packages into its own directory, which is outside of WSL.

PR #52, opened 2017-07-04
add Base16 color scheme

tnguyen14:
This was generated using https://github.com/mbadolato/iTerm2-Color-Schemes/blob/master/tools/xrdb2Xresources.py
from a xrdb file, which was generated by using https://github.com/mbadolato/iTerm2-Color-Schemes/blob/master/tools/iterm2xrdb
on https://github.com/chriskempson/base16-iterm2/blob/master/base16-default.dark.256.itermcolors

This is to address the first issue raised in #47.

The changes to the `package.json`s are done automatically by npm (I'm using npm v5). If you'd like to not include those changes, please let me know.

I've tested this out and it seems to work, however, it seems that terminus does not support 256 colors, because colors 17-21 appear as black/ blue (https://github.com/chriskempson/base16-shell#troubleshooting)

### DIFF
--- a/terminus-terminal/package.json
+++ b/terminus-terminal/package.json
@@ -2,7 +2,9 @@
   "name": "terminus-terminal",
   "version": "1.0.0-alpha.14",
   "description": "Terminus' terminal emulation core",
-  "keywords": ["terminus-plugin"],
+  "keywords": [
+    "terminus-plugin"
+  ],
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
   "scripts": {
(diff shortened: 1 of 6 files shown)


terminus-terminal/package.json
```diff
@@ -34,12 +36,12 @@
     "rxjs": "5.3.0"
   },
   "dependencies": {
-    "font-manager": "0.2.2",
```
Eugeny: Please revert changes that affect dependency versions. Changes to formatting are fine though!

Eugeny:
And could you please also post the 256 color support bug as a separate issue? Thanks.

tnguyen14:
I've removed the package version changes, and opened #59.

PR #64, opened 2017-07-05
I'd like to improve the design of the github page

darrionramdin:
I'm currently practicing my web design skills and would like to give the github make a more modern look as well as make it responsive it. Can I contribute that way?

Thanks.
Attached is an image of how it looks completed. The video plays automatically as how it did in the original website.

terminusafter


princemaple:
I think the current page, although not very responsive around the screenshot, is more informative. And I prefer the looks of it. (VERY PERSONAL OPINION)

tnguyen14:
I think the current page has styles and colors that are consistent with the design of the app itself. Not sure if that's necessary, but I think it's nice.

Eugeny:
I'm torn by rejecting someone's hard work like this, but I agree with the people above. 
But well, this is the thing with unsolicited redesigns :)

PR #97, opened 2017-07-06
update Base16 Default Dark theme to 256 colors

tnguyen14:
This is a follow up on #52. Apparently I converted the theme that did not  support 256 colors.

This change will update it to make full use of 256 colors.

### DIFF
--- a/terminus-community-color-schemes/schemes/Base16 Default Dark
+++ b/terminus-community-color-schemes/schemes/Base16 Default Dark
@@ -12,27 +12,27 @@
 !
 ! Red
 *.color1:      #ab4642
-*.color9:      #dc9656
+*.color9:      #ab4642
 !
 ! Green
 *.color2:      #a1b56c
-*.color10:     #282828
+*.color10:     #a1b56c
 !
 ! Yellow
 *.color3:      #f7ca88
-*.color11:     #383838
+*.color11:     #f7ca88
 !
 ! Blue
 *.color4:      #7cafc2
-*.color12:     #b8b8b8
+*.color12:     #7cafc2
 !
 ! Magenta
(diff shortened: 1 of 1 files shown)


tnguyen14:
I just downloaded alpha 20 to try this out. It seems that the colors are still the old ones.

Is there a way to verify that I have the right version running? And also a way to make sure the styles are not cached somehow?

tnguyen14:
@Eugeny do you know why the colors might not be updated? Is there a way I can bust the cache (if that was the issue)?

Eugeny:
@tnguyen14 try selecting a different color scheme and then Base16 again.

tnguyen14:
Yup I tried that, but doesn't seem to work.

Here's the screenshot of what I'm seeing. On the second line, the "G" color is actually grey:

base16-theme

tnguyen14:
@Eugeny I know you're probably incredibly busy. Just wanted to see if you have any suggestion for how I can debug this?

Eugeny:
@tnguyen14 I'm sorry, I've missed the notification for this thread. Are the colors in your `config.yaml` correct when you select the theme from the list? Can you try the colorpicker (from Dev Tools on the Settings page) to check that the color is actually wrong?

tnguyen14:
The devtools show the color for `c10` as `rgb(40, 40, 40)`, which is the old (`#2828281`) color. The new one should be `rgb(161, 181, 108)`.

Where do I ese the contents of `config.yaml`?

tnguyen14:
Interestingly, when I go to the "Plugins" tab, I see 2 `community-colorscheme` plugins, one which I could delete. So I proceeded to delete that, restart the app, and the color is coming through correctly.

PR #167, opened 2017-08-07
Start an interactive logon shell for Git Bash

### LINKED ISSUE #105: Missing options to cygwin bash
Just tried Terminus 1.0.0 alpha 19 on Windows 10 with Cygwin bash.
What I've noticed is that the PATH is not properly set (e.g. /usr/bin/ is not included in the PATH).
Comparing to the cygwin.bat terminus just run `bash`, while the bat file runs `bash --login -i`

koraktor:
Provide additional arguments to `bash.exe` to get an interactive login shell. This ensures e.g. `.profile` and `.bash_profile` are sourced. As there’s no way to have an existing session under Windows, `--login` is mandatory. Each bash session must be started from scratch.

Fixes #105


Eugeny:
Thanks! :beer:

PR #241, opened 2017-11-21
use base16-xresources

tnguyen14:
I realized that the author of this theme maintains an XResources file, so I figured I'd just copy from it directly.

### DIFF
--- a/terminus-community-color-schemes/schemes/Base16 Default Dark
+++ b/terminus-community-color-schemes/schemes/Base16 Default Dark
@@ -1,44 +1,54 @@
-!
-! Generated with :
-! XRDB2Xreources.py
-!
-*.foreground:  #d8d8d8
-*.background:  #181818
-*.cursorColor: #d8d8d8
-!
-! Black
-*.color0:      #181818
-*.color8:      #585858
-!
-! Red
-*.color1:      #ab4642
-*.color9:      #ab4642
-!
-! Green
-*.color2:      #a1b56c
-*.color10:     #a1b56c
-!
-! Yellow
-*.color3:      #f7ca88
-*.color11:     #f7ca88
-!
-! Blue
-*.color4:      #7cafc2
-*.color12:     #7cafc2
-!
-! Magenta
-*.color5:      #ba8baf
-*.color13:     #ba8baf
-!
-! Cyan
-*.color6:      #86c1b9
-*.color14:     #86c1b9
-!
-! White
-*.color7:      #d8d8d8
-*.color15:     #f8f8f8
-!
-! Bold, Italic, Underline
-*.colorBD:     #d8d8d8
-!*.colorIT:
-!*.colorUL:
+! Base16 Default Dark
+! Scheme: Chris Kempson (http://chriskempson.com)
+
+#define base00 #181818
+#define base01 #282828
+#define base02 #383838
+#define base03 #585858
+#define base04 #b8b8b8
+#define base05 #d8d8d8
+#define base06 #e8e8e8
+#define base07 #f8f8f8
+#define base08 #ab4642
+#define base09 #dc9656
+#define base0A #f7ca88
+#define base0B #a1b56c
+#define base0C #86c1b9
+#define base0D #7cafc2
+#define base0E #ba8baf
+#define base0F #a16946
+
+*.foreground:   base05
+#ifdef background_opacity
+*.background:   [background_opacity]base00
+#else
+*.background:   base00
+#endif
+*.cursorColor:  base05
+
+*.color0:       base00
+*.color1:       base08
+*.color2:       base0B
+*.color3:       base0A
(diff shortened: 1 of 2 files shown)


tnguyen14:
@Eugeny do you know why the appveyor build might have failed?

Eugeny:
Have you tested the changes locally?

tnguyen14:
No I have not, I didn't expect the changes to have any build impact, but I could be wrong as the last time I contributed was months ago. I will try to run the test locally tonight.

Eugeny:
The problem is that the Xresources parser in Terminus can't handle variables. (the failed build isn't related to your pull request)

tnguyen14:
Thanks. Could you point me to where the Xresources parser is in Terminus? I tried to look around briefly, but it wasn't obvious to me.

tnguyen14:
If we can't change the parser, I think I can manually translate it to what the values should be. Upon inspection, it looks like the new one just add colors 17-21, and the old one has `colorBD` defined.

Eugeny:
Link to the parser: https://github.com/Eugeny/terminus/blob/master/terminus-community-color-schemes/src/colorSchemes.ts

tnguyen14:
Thanks for the reference. Do you know how the colors are used? I see that it is only configured to accepts colors from 0 to 15. If I add colors 16-21, will the renderer pick them up?

Eugeny:
I've just updated the parser to include any colors beyond 16. Could you please try it out with your theme?

tnguyen14:
I'm trying to run with the latest master - after `npm run build` and `npm start`, when the app loads, I see the following error on the devtools console

```
ERROR Error: Uncaught (in promise): [object Object]
    at resolvePromise (/Users/tnguyen541/github/Eugeny/terminus/app/node_modules/zone.js/dist/zone.js:770)
    at /Users/tnguyen541/github/Eugeny/terminus/app/node_modules/zone.js/dist/zone.js:696
    at rejected (css-base.js:75)
    at ZoneDelegate.invoke (/Users/tnguyen541/github/Eugeny/terminus/app/node_modules/zone.js/dist/zone.js:391)
    at Object.onInvoke (ng_zone.ts:296)
    at ZoneDelegate.invoke (/Users/tnguyen541/github/Eugeny/terminus/app/node_modules/zone.js/dist/zone.js:390)
    at Zone.run (/Users/tnguyen541/github/Eugeny/terminus/app/node_modules/zone.js/dist/zone.js:141)
    at /Users/tnguyen541/github/Eugeny/terminus/app/node_modules/zone.js/dist/zone.js:818
    at ZoneDelegate.invokeTask (/Users/tnguyen541/github/Eugeny/terminus/app/node_modules/zone.js/dist/zone.js:424)
    at Object.onInvokeTask (ng_zone.ts:288)
```

Upon clicking the `+` button to create a new shell, nothing happens and the same error appears.

I know the error doesn't seem to be very helpful. I was wondering if you see any similar error on your end?

tnguyen14:
I also see the following error in the console, but I'm not sure if it's related

```
> cross-env DEV=1 electron --js-flags='--ignition' app --debug

Host startup: 128ms
2017-11-28 00:27:43.398 Electron Helper[97949:4646066] Couldn't set selectedTextBackgroundColor from default ()
info: v 1.0.0-alpha.1
info: Update available: 1.0.0-alpha.37
```

tnguyen14:
@Eugeny do you know how I could debug this? The stack trace isn't very helpful. I've tried re-running `npm run build` but that doesn't do anything.

Eugeny:
Sorry for the delay. You can enable "stop on exceptions" in the DevTools, that should get you to the point it crashes at.

tnguyen14:
Thanks for the tip!

The `[object Object]` above is as follows:

```js
{
  "time": 1512172203,
  "number": 302,
  "lines": [
    "invalid option: bell-on-alert"
  ],
  "error": true
}
```

It is of type `TMuxBlock`.

Eugeny:
Which tmux version are you running?

tnguyen14:
```sh
:; tmux -V
tmux 2.6
```

tnguyen14:
@Eugeny is there a different tmux version I should try with?

tnguyen14:
@Eugeny Do you have any suggestion here on how I can continue to debug this?

Eugeny:
Please post this tmux error as a separate bug and switch to Screen or disable session persistence.

tnguyen14:
I've updated the script to process `#define` lines as well.

So the theme as it is will work. The only thing the processor doesn't know about is `#ifdef` and `[background_opacity]`. However, it will read `*.background` twice, and so the second value will override the first, so it works.

Eugeny:
Looks good now!

tnguyen14:
Thanks! Please let me know when this is released as a version, I will download it on my Windows machine and test it there.

PR #252, opened 2017-12-01
Include all build tools in project's dependencies

tnguyen14:
This change allows for the deps install step to be run without requiring a global package to be installed.

What this means is that when starting the project from scratch, the developer can run `npm install` to install the dependencies, then the `install-deps.js` step will use the locally installed `yarn` bin to install the inner dependencies.

This is just a suggestion - if you think it's a good idea, then I will go ahead and update the `HACKING.md` guide to reflect this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "node-gyp": "^3.6.2",
     "node-sass": "^4.5.3",
     "npmlog": "4.1.0",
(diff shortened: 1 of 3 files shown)


Eugeny:
Looks good! Feel free to update the doc too.a

PR #261, opened 2017-12-14
feat(tab): enable cycle tab selection

### LINKED ISSUE #258: Cycle through tab selection
Please feel freely close if this is dupe or already available. Is there way to configure single hotkey to cycle tabs through?

kwonoj:
- closes #258 

I took a stab myself for this feature, while I also have relatively high confidence PR is not feasible to be accepted. 

This PR exposes one more settings values to allow `cycle` tab selection, either next / previous tab hotkey will select to last / first tab if user specifies to.

point of consideration
1. may possible to control this behavior via plugin instead of core?
: I couldn't find relative interface support, but maybe I am lack of knowledge simply.

2. Is it feasible to expose under application settings, under `appearance`? 
: It isn't actually appearance but more about behavior, while current settings doesn't have specific category. Hotkey provider is managing assigned key combination, so picked current one instead.

3. Would it be better to have checkbox instead of radio?
: this may not be big deal (and could reuse all styles), but still this could be checkbox for boolean flag.

### DIFF
--- a/terminus-core/src/configDefaults.yaml
+++ b/terminus-core/src/configDefaults.yaml
@@ -3,6 +3,7 @@ appearance:
   dockScreen: current
   dockFill: 50
   tabsLocation: top
+  cycleTabs: true
   theme: Standard
   frame: thin
   css: '/* * { color: blue !important; } */'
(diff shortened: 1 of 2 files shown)


Eugeny:
Thanks for your contribution! 🍺 
A fine addition, but I think that it should actually be the default behaviour! Please set it to `true` in the config defaults.
This is a very minor setting that doesn't need GUI, I'll just add it to the config docs.

kwonoj:
@Eugeny cool! updated PR as suggested.

PR #285, opened 2018-02-02
Fixed typo in SSH client's configuration window.

VaultVulp:
It looks like somebody misspelled "port", thus blocking the users from accessing SSH-servers that listen on non-default port.


Eugeny:
Thank you!

PR #357, opened 2018-06-04
Update popup to get rid of messy process identifiers

peterfraedrich:
`children[0].process` would show process identifiers like `s001` instead of a process name on an idling terminal. Changed close confirm message so that this doesn't happen.

I like the idea of showing the process when confirming, but I would change it to this until there's a way to reliably produce a "pretty" process name.


peterfraedrich:
I'm occasionally getting these when closing idle tabs, when when its just sitting there at a prompt.

snip20180605_3

Eugeny:
Is there actually a process running, other than shell? This looks like a `ps` parsing bug actually

peterfraedrich:
No, the only process running is `zsh`

PR #366, opened 2018-06-23
Allow modal dismiss via click

levrik:
I just tried out Terminus and I really like it.
But there was this thing that really annoyed me so I fixed it 🙂


Eugeny:
Thank you!

PR #402, opened 2018-08-29
Add login scripts support

### LINKED ISSUE #344: Feature Request: XShell liked login script
I like terminus. I used to use XShell as the primary ssh client. And the most missing feature of terminus is the ability to execute some script after login. Without this feature, I cannot automate my work.
The login script in XShell is like expect program. It wait for some string, and then send a text to the console.

Domain:
Support login scripts. Fix #344


Eugeny:
Thank you for your contribution! Could you please adjust braces to follow the general Terminus style (opening brace on the same line)?

Domain:
Done

Issue #374, opened 2018-07-09
Password tab remained deselected after clicking on an ssh connection.


vishydewangan:
Password tab remained deselected after clicking on an ssh connection. It required to click on on this tab to enter password. 

Version: 1.0.0-alpha.48
Platform: win32 10.0.16299

Issue #403, opened 2018-08-29
better handling for large numbers of ssh connections


kbjr:
I have something like 200 SSH connections saved at any given time.

First, when you have a ton of connections, the UI actually breaks. I had to apply some custom CSS to make the connection selector actually scroll instead of flowing off into nowhere.

#### My hacky fix for large SSH connection list:

```css
input[placeholder^='Quick connect'] + div.list-group.mt-3,
.list-group.mt-3.ng-star-inserted + div.list-group.mt-3 {
    max-height: 800px;
    overflow-y: auto;
}
```

Second, more of a nice to have thing, it would be really nice to have some way of quickly getting to the connection that I want, rather then having to scroll through a large list looking for it, like a fuzzy search/filter option.

Domain:
I want to organize the connections in folders. But I am not familiar with angular. I want something like this:
[https://jmouriz.github.io/angular-material-multilevel-menu/demo/demo.html](https://jmouriz.github.io/angular-material-multilevel-menu/demo/demo.html)

Eugeny:
@kbjr how does it break exactly? The modal becomes scrollable if there isn't enough space in my case

kbjr:
It didn't for me, I'll try and get some screenshots tonight

kbjr:
![image](https://user-images.githubusercontent.com/195127/45183205-4ac0c600-b1d8-11e8-85a1-5f0a9e52b2fd.png)

Okay, I guess technically it is scrollable, in the sense that the entire modal can be moved around in view, but it does clearly hang off the edge of the UI and looks odd.

The custom CSS I added actually scrolls within the modal

![image](https://user-images.githubusercontent.com/195127/45183287-88bdea00-b1d8-11e8-9239-8f55c934cf64.png)

Eugeny:
Fixed - the latest code also lets you filter the entries using the "quick connect" field

kbjr:
@Eugeny Cool, I see that an alpha.55 was was tagged, but there's no release packages. Are those coming soon? Would love to actually have this feature

Eugeny:
@kbjr my certificate recently expired so I'm reconfiguring the build back to unsigned. Binaries will be up soon!

Eugeny:
It's out https://github.com/Eugeny/terminus/releases/tag/v1.0.0-alpha.55

PR #409, opened 2018-09-04
Organize the connections in folders. 

Domain:
Organize the connections in folders. 
Filter connections in quick connect.
Need more custom CSS to handle long connection list.
See also #403


Domain:
This also fix #374

PR #418, opened 2018-09-13
Save window state on close.

BBJip:
This pull request fixes the issue of unsaved window states. If you resize a window, maximize, close and reopen it, the window will now restore itself in maximized window mode, while also remembering the previous window size and position of the unmaximized window.

Before the patch, if you closed and reopened a maximized Terminus, it led to ugly bars on the sides of the window (at least if you were running Terminus on Windows):
previouswindowrestore


Eugeny:
Thank you for the contribution! :beer:

BBJip:
Thanks for making such a great terminal app!

Issue #534, opened 2018-12-04
terminus self-built doesn't start because of odd squirrel handling


Drachenkaetzchen:
Prerequisites:
- Build from source, no errors

Symptom:
- Run `npm start`
- Observe the following message: `Angular bootstrapping error: Error: Could not call remote function 'checkForUpdates'. Check that the function signature is correct. Underlying error: Can not find Squirrel`

The electron autoupdater API unfortunately doesn't expose if Squirrel is available or not, so we need to rely on catching the error. Due to my limited TypeScript/Electron knowledge, I'm not sure if that's the best way, but I'll submit in a PR in a few minutes anyways.

Drachenkaetzchen:
Also I noticed that there's a try/catch around `setFeedURL()`, and I'm not sure why that's there. `setFeedURL()` doesn't actually check if the URL is there.

Eugeny:
Right-o, I totally forgot about `checkForUpdates`. `setFeedURL` fails without Squirrel as well

Eugeny:
Fixed in 9fd69f668a48ad1d02ad338a4ac45c900234d64f

PR #535, opened 2018-12-04
Bugfix for issue #534: Catch any errors occurring during checkForUpda…

Drachenkaetzchen:
…tes() to allow App to start up even if Squirrel is not available


Eugeny:
Could you change it to check `this.isSquirrel` instead? I'm already detecting Squirrel presence at updater.service.ts:24

PR #633, opened 2019-01-29
Update windows-swca dependency

sylveon:


app/tsconfig.json
```diff
@@ -9,7 +9,6 @@
         "emitDecoratorMetadata": true,
         "experimentalDecorators": true,
         "sourceMap": true,
-        "noUnusedParameters": true,
```
Eugeny: Any particular reason?

sylveon: It's in duplicate. The same entry shows up a few lines below.

Eugeny: ah, gotcha

PR #647, opened 2019-02-05
Change close behavior on MacOS

martykan:
On MacOS, the close button isn't actually supposed to quit the app, it should only hide it.
It should be changed so it's consisted with the platform


martykan:
OK, I see, I didn't take multiple windows into account.

PR #766, opened 2019-04-16
Create close pane hotkey.

Mezner:
I wanted to have an ability to close panes as easily as creating them, so I created a simple shortcut for it.


Eugeny:
Thanks a lot!

Mezner:
Thanks for the merge. Is there a way I can self-release a mac dmg so I can run this without engaging another terminal? How often do you cut releases?

Eugeny:
Generally, just every now and then. I've just started the Alpha 74 build, but in the future you can follow the steps here: https://github.com/Eugeny/terminus/blob/master/.travis.yml#L29

Eugeny:
@all-contributors add @mezner for code

PR #1043, opened 2019-06-08
Remove name in package.json

NieLeben:
Removed name because it's a monorepo and it conflicts with https://www.npmjs.com/package/term


Eugeny:
@allcontributors add @NieLeben for code

PR #1047, opened 2019-06-09
Set overflow on content instead of screen in settings

IgnusG:
Resolves https://github.com/Eugeny/terminus/issues/1040


Eugeny:
@allcontributors add @IgnusG for code

PR #1138, opened 2019-06-28
Bugfix - Window starts outside of display bounds

ehwarren:
### Description
When running multiple displays it's easy to end up with a terminus window that starts up with bounds outside of any of the currently utilized displays.

https://github.com/Eugeny/terminus/issues/844

### Steps to reproduce
1. Launch instance of terminus on a setup utilizing multiple displays
2. Move the terminus window to a *non-primary* display (Now referred to as *display2*).
3. Close the terminus window.
4. Turn off *display2*.
5. Start a new instance of terminus.
6. The instance will launch, and will be visible in the taskbar and alt-tab menus, but is not possible to be brought over to the main display.

### System configuration
**OS**: Windows 10 Pro - 10.0.17763 Build 17763
**Terminus Version**: 1.0.0-alpha.1

### Proposed solution
When the application first opens, we use Electrons functionality to figure out which screen is closest to the previous top/left position of our window.

We then check to see if the previously saved window bounds configuration overlaps with the closest screen. If they do, then we know the window is visible and we don't do anything. If they don't overlap, we cannot see the window; in this case we simply center the window in whatever screen was closest to the screen we were previously on.

### Notes
In this proposed solution I use whatever window width/height was last saved in the configuration. Perhaps it would be better to revert back to the default width/height? Let me know what you think.


ehwarren:
I noticed that the build failed on my local machine until I manually ran 

`yarn build:typings`

perhaps that's why this build is failing in travis?

Eugeny:
Thanks a lot! I've also just set `build:typings` to be auto-run by `build`. CI might be flaky right now as I'm in process of moving to Azure Pipelines.

Eugeny:
@allcontributors add @ehwarren for code

PR #1277, opened 2019-07-20
Add a newline after banners

ThePuzzlemaker:
This is to prevent ASCII-art style banners from getting messed up by the offset of the 'Banner: ' text such as in [this image](https://i.imgur.com/K78mEus.png).

Apologies if this code doesn't work, I'm not exactly good with TypeScript.


Eugeny:
Thanks a lot!
@allcontributors add @ThePuzzlemaker for code

PR #1304, opened 2019-07-24
Allow configuration of new tab CWD behaviour

### LINKED ISSUE #1239: New tab always opens in the same directory as previous tab
I have a working directory set in Terminus' settings but everytime I open a new tab it opens in the same directory as the previous tab.

ehwarren:
Fixes #1239 
Added option in "shell" section of settings. Option to always use working directory defaults to false.


Eugeny:
Looks good - just check the lint please

ehwarren:
Not a fan of semicolons eh... I guess I can live with that ;)

PR #1385, opened 2019-08-13
Add Linux incompatibility notice

KsRyY:
This PR adds a notice for Linux users that `sudo` does not work on Terminus >=1.0.78(Alpha 78).


Eugeny:
sudo execution has now been fixed

PR #1428, opened 2019-08-29
Fix btn color in paper theme

yfwz100:
Fix btn color in paper theme.


Eugeny:
Thanks! 
@allcontributors add @yfwz100 for code

PR #1431, opened 2019-08-29
Check if HOME env var doesn't already exist (Windows)

### LINKED ISSUE #1378: Terminus replaces $HOME environment variable when opening
When I open a new terminal in Terminus, my $HOME environment variable is getting replaced to `C:\Users\USERNAME`. While it's nice to have that as a default, it's really annoying when I always have to refresh environment variables to get the one that I've set.

jack1142:
Fixes #1378

### DIFF
--- a/app/src/entry.ts
+++ b/app/src/entry.ts
@@ -18,7 +18,7 @@ location.hash = ''
 
 ;(process as any).enablePromiseAPI = true
 
-if (process.platform === 'win32') {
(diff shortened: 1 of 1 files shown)


jack1142:
It looks like the failing build isn't connected to my changes but correct me if I'm wrong.

Eugeny:
Thanks!

@allcontributors add @jack1142 for code

Issue #1554, opened 2019-09-26
Can't close search pane


Leonorus:
Can't find the way for close search pane after opening it for the panel. 
Also the preferences for each search not personal. If I toggle the "Whole word" selector on one he toggles one all another.
![search-selectors](https://user-images.githubusercontent.com/4058350/65679202-4bcfca00-e05d-11e9-80cb-2a9e92f8fd48.gif)

PR #1615, opened 2019-10-14
Issue #1554 - Make search options not static/global

ckaczor:
Secondary part of #1554 - make search options not static/global


Eugeny:
Thanks a lot!

@allcontributors add @ckaczor for code

PR #1929, opened 2019-12-29
Fix broken links in core README

LeSeulArtichaut:


Eugeny:
Thanks, but this README is actually used to generate docs here: http://ajenti.org/terminus-docs/ and the links are correct

LeSeulArtichaut:
Got it. Sorry for this misunderstanding

PR #2081, opened 2020-02-02
Fix: save plugins to path `UserData` for fit portable mode

CyrilTaylor:


Eugeny:
What is Sentry using `userData` for? If it's just crash info, it's actually better to keep it local to the PC instead of portable data.

CyrilTaylor:
> What is Sentry using `userData` for? If it's just crash info, it's actually better to keep it local to the PC instead of portable data.

terminus-data\sentry\scope_v2.json

Eugeny:
Yeah - it's definitely OK to keep it local. I'll merge your first commit and close this PR

CyrilTaylor:
But I think the user who like portable mode want to keep local pure.

Eugeny:
Good point. Can you move it to its own file then? (portable.ts, to be included before sentry.ts)

Eugeny:
Thanks!

PR #2112, opened 2020-02-12
improve the launch performance of portable

CyrilTaylor:
For portable, use the unpack binary.

|launch time|cold launch|hot launch|
|-|-|-|
|old|>14s|>10s|
|new|<5s|<1s|

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,10 @@
 
 ---
 
+# Portable
+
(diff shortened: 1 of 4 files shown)


Eugeny:
Looks good :+1:

I'll swap the target for zip to avoid users needing to have 7zip to unpack it.

PR #2148, opened 2020-02-19
resolves #1988

### LINKED ISSUE #1988: home and end key does not work
Version: 1.0.98
Platform: win32 10.0.18362
Plugins: none

I create a ssh connection by **[Settings]->[SSH]**, when i open it through **[Quick connect]**,  "home" and "end" key cannot work, when i open a new tab through **[New terminal]**, it work correctly.

nstefanou:
This should fix the home/end hotkey problems on ssh terminals


Eugeny:
Any idea why the behaviour is different on SSH vs. local terminals?

nstefanou:
> Any idea why the behaviour is different on SSH vs. local terminals?

I'm no expert, but it may have to do with the TERM variable being wrong, or because of the default shell on the remote session
This fix is what worked for me on my environment

PR #2241, opened 2020-03-11
Have multiple recent connections in history instead of just one

nstefanou:
Hi,  

I thought this would be useful since a user may want to connect quickly to multiple machines without having to save them, due to IP address changes or for any other reason.  
I put an arbitrary number of 5 recent connections, but I think it'd be better if it was a configuration option.  
It's just a suggestion, you can totally disregard this PR if you don't agree with it.  

Cheers


Eugeny:
Looking good! :+1:

PR #2364, opened 2020-04-08
Add an author and license fields

orin220444:


Eugeny:
Thanks! @allcontributors add @orin220444 for code

PR #2437, opened 2020-04-20
Add dock always on top setting

### LINKED ISSUE #1894: Terminus Continues To Overlay When Clicked Off
Version: 1.0.92
Platform: win32 10.0.18362
Plugins: none

Hey guys the Terminus window will stay above all other windows when I click off. I have a setup where I press Alt + Space to open it which takes up the upper portion of my screen. Then when I click off to say Chrome or Sublime it still stays on top which is quite annoying. Only way to hide it again is to press Alt + Space. Also when it is setup like this, Terminus doesn't show in the Windows Dock as an open app, it looks like it is closed so when you click the icon in the dock, it opens another session of the app.

Goobles:
Adds a dock "always on top" toggle.

This possibly fixes #1894

This also tweaks the behavior of the toggle-window shortcut so it behaves as expected
(brings unfocused docked window to top, but still closes unfocused docked window that is already on top),

### DIFF
--- a/app/lib/window.ts
+++ b/app/lib/window.ts
@@ -193,8 +193,13 @@ export class Window {
                     this.window.focus()
                 })
             } else {
-                // docked, visible
-                this.window.hide()
+                if (this.configStore.appearance?.dockAlwaysOnTop) {           
+                    // docked, visible, on top
+                    this.window.hide()
+                } else {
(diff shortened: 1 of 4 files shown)


Eugeny:
Looks good - thanks a ton! 
@allcontributors add @Goobles for code

PR #2601, opened 2020-05-20
Jmb/features

JonathanBeverley:
Fix for #350. "postinstall" script in terminus-ssh/package.json only works in Windows, but that's fine, because we only need the pagent connector on Windows.


Eugeny:
Sorry for taking so long :) 

@allcontributors please add @JonathanBeverley for code

PR #2824, opened 2020-07-27
Fixed SSH Keys Permission Denied

Dimitory:
The issue reproduce for domain users ( domain\username, actually code uses only username )


Eugeny:
Nice :+1: 

@allcontributors add @dimitory for code

PR #3055, opened 2020-10-08
Add possibility to configure always enabled regex

matishadow:
Hello,

When I want to copy something from terminal (like a part of some output) I always use regex in search.
I believe many people also does this. 

I thought it'd be really nice to be able to configure regex in search so it's always enabled.


Eugeny:
Would you consider changing it to just remember the search options instead? I think that's more flexible than just having a setting for this one option.

matishadow:
I adjusted it so the search options are saved without settings.

Eugeny:
:+1:

@allcontributors please add @matishadow for code

PR #3072, opened 2020-10-14
feat: Add copy button to ssh connection

zend:
I have a lot of SSH connection profiles, but by now I need to add them one by one. 

But in fact, all the profiles have some common fields like username, password, private key and even login scripts. 

This PR adds a copy button to add a new connection based on the existing connection profile.


Eugeny:
Amazing, thanks! @allcontributors please add @zend for code.

PR #3154, opened 2020-11-22
fix: open in here on new window

FreddyZeng:
open in here is not work on PC windows, I fixed it on New window. And keep old logic work.


ColinEspinas:
@FreddyZeng Can you fix the linting error or do you want someone else to do it? 
I think this is a very important fix and it would be really good if it was merged.

Thanks, have a great day!

PR #3303, opened 2021-01-06
add baudrate 1500000

frauhottelmann:
1500000bps is needed for single board computers with RK3399 CPUs

I hope this change is enough 😄


Eugeny:
Thanks! @allcontributors please add @frauhottelmann for code

PR #3362, opened 2021-01-21
Removes unnecessary asterisks from README

mi544:


mi544:
Not sure if those asterisks are there for a specific reason, but I don't remember seeing them in previous README files

Eugeny:
Thanks!

PR #3365, opened 2021-01-24
Yarpm

Milo123459:
Hello 👋!

I've added yarpm, a tool that detects if you are using npm or yarn to run a script and uses that CLI. Pretty cool!

Hope you have a great day!

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,14 +57,15 @@
     "val-loader": "2.1.1",
     "webpack": "^5.11.0",
     "webpack-cli": "^4.2.0",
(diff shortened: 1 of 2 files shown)


Eugeny:
Thanks, I appreciate the effort but Terminus strictly requires Yarn.

Milo123459:
Why is npm run being called in the package.json then?

Eugeny:
Purely historical reasons - Terminus has used NPM previously. Although `yarn run` vs `npm run` would have no functional difference here, feel free to correct :)

PR #3489, opened 2021-02-22
Add Nord theme

VectorKappa:


Eugeny:
@allcontributors please add @VectorKappa for design

PR #3543, opened 2021-03-04
FIX #3520

### LINKED ISSUE #3520: SSH Host with same IP and different Port can not save password separately
Version 1.0.134 (1.0.134)   MacOS Version

I have two accounts on the same host with different password.   e.g    root@1.2..3.4:22   root@1.2.3.4:10022
I created two entries in SSH manager,  but I found that the two entries shares the same saved password.

Maybe the unique key for an SSH entry should be   user+IP+port 

This case is common when you use autossh as a reverse tunnel for servers without public IP address

iamchating:
Hi, I have a quick & dirty fix for this issue, maybe it works.

I am sorry I only "test" it by my brain :)

### DIFF
--- a/terminus-ssh/src/services/passwordStorage.service.ts
+++ b/terminus-ssh/src/services/passwordStorage.service.ts
@@ -4,15 +4,27 @@ import * as keytar from 'keytar'
 
 @Injectable({ providedIn: 'root' })
 export class PasswordStorageService {
-    async savePassword (connection: SSHConnection, password: string): Promise<void> {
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks! Can you adjust it to only include the port if it's not 22? That way the old saved passwords won't be "lost".

Eugeny:
:+1: just check the linter messages and I'll merge

iamchating:
Will open a new pull req.

iamchating:
I am not familiar with github, and I have just pushed some fix on lint errors but not shown here

Eugeny:
No problem!

PR #3554, opened 2021-03-05
FIX: #3520

iamchating:


iamchating:
It seems that there is no lint errors in log, but the error in installing deps

Eugeny:
@allcontributors please add @iamchating for code

PR #3952, opened 2021-06-05
Add Iceberg color scheme

lambdalisue:
Thanks for the project. I'd love to use the Iceberg color scheme so I've added it. Please review.

https://cocopon.github.io/iceberg.vim/
https://github.com/cocopon/hyper-iceberg/blob/master/index.js


Eugeny:
Thanks! @allcontributors please add @lambdalisue for design

PR #4077, opened 2021-06-29
Update "maximized" icon to be more consistent with standard Windows UI

bdr99:
Thanks for fixing my issue https://github.com/Eugeny/terminus/issues/2712! I noticed that the new icon has an additional horizontal bar which makes it look quite different from most Windows applications. This PR removes the bar to make the icon more closely match the way almost all Windows apps look.

Before:
image

After:
image

Example from Windows 10 File Explorer:
image


Eugeny:
Thanks! @allcontributors please add @bdr99 for design

PR #4132, opened 2021-07-05
Update header anchor link

logicmachine123:
Update header anchor link to it's new package name "tabby", not updating this anchor link results in a 404 when clicking in the BEAM ME UP button.


Eugeny:
Thanks! @allcontributors please add @logicmachine123 for docs

PR #4168, opened 2021-07-08
add Ubuntu 20.04 compile requeriments

cypherbits:


Eugeny:
Thanks for contributing. What was cmake required for?

cypherbits:
> Thanks for contributing. What was cmake required for?

When executing build-native script.

Eugeny:
Got it. I've merged it together with the Debian list, since these things would be required there too.

@allcontributors please add @cypherbits for docs

PR #4215, opened 2021-07-16
Correctly name profiles for shells outside /usr

KingMob:
Shells outside `/usr` get named incorrectly. E.g., if I add `/usr/local/bin/fish` to `/etc/shells`, when Tabby starts up, it think the profile name should be "local".


Eugeny:
Thanks! @allcontributors please add @KingMob for code

PR #4363, opened 2021-08-06
Profile filter per search term instead of the entire search input

al-wi:
Split up the search input of the profile filter, and check if every search term is present in a profile name. This makes searching in a long list of profiles easier.


Eugeny:
Thanks! @allcontributors please add @al-wi for code

PR #4545, opened 2021-09-06
Update README.md in Korean

BoYeonJang:
As a user who frequently uses tabby as an SSH client, I translated README.md hoping that many Koreans would use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 
 ----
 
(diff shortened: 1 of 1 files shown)


Eugeny:
Hi, I appreciate the effort that went into this, but I can't just replace the README with a Korean version as there are users from other countries too :)

PR #4548, opened 2021-09-07
Created a READMD.ko-KR.md translated into Korean

BoYeonJang:
I was wrong. It created a READMD.ko-KR file translated into Korean. Can you add a Korean shortcut link?


Eugeny:
@allcontributors please add @BoYeonJang for docs

PR #4959, opened 2021-11-19
Add description to port forwarding and autofill inputs with the deleted port

highfredo:
Hi!

I have been using it for a few days and I think it would be great to be able to put descriptions to the port forwardings, or at least for me it would be very useful to remember why port 3593 points to 7234 😛 

I have also made that when a port is deleted, this information remains in the creation form, in case you wanted to edit it or have accidentally deleted it


Eugeny:
This looks great, thanks a lot! @allcontributors please add @highfredo for code

PR #4962, opened 2021-11-21
Deepclone profile before passing it to the edit configuration component and fix uncaught error on modal dismiss

highfredo:
Hello again!

When editing the configuration of the ssh profiles there were some options that were saved without clicking the save button (eg ssh port forwarding). It is fixed by doing a deepClone of the profile instead of Object.assign.

Also catch the error when the modal is dismissed.


Eugeny:
Thanks! I've flipped the catch around in 2baede467511a18daf53f8b160c394ffbd2c93f4 though to avoid swallowing errors that are not caused by the modal dismissal itself

PR #4992, opened 2021-11-25
Fix hotkey setter to handle backticks and tildes properly

### LINKED ISSUE #4006: Hotkey setter treats backtick as tilde.
Using a backtick in hotkeys doesn't work because it gets converted into a tilde.

For example, if you want `` Ctrl + ` `` as the toggle key, it will be registered as `Ctrl + ~`.

Editing the config file directly works as a workaround.

NessunKim:
This fixes #4006.


Eugeny:
Thanks! That's much easier than I feared it was going to be :D 
@allcontributors please add @NessunKim for code

PR #5142, opened 2021-12-13
Add translation workflow

songdaochuanshu:
A translation workflow has been added to ensure that README is updated Synchronize other translations after 🤖


Eugeny:
Hi, thanks for the PR! :wave: I appreciate the effort you've put into this, but it's basically the same as the user translating the page using the browser's built-in translation, so I won't be merging this.

Issue #4954, opened 2021-11-19
Paper theme: unreadable text in SFTP and top bar


jerryuhoo:
![image](https://user-images.githubusercontent.com/24245709/142572294-da9c08d8-9f56-47a7-9956-d273b7cf3ddb.png)
![image](https://user-images.githubusercontent.com/24245709/142572411-b79d6790-4371-48c7-a97e-9692f359ba95.png)
I use the latest version (1.0.164) of Tabby. I used Standard theme before. When I switched to Paper theme, the background of SFTP panel is still the dark color of Standard theme.

SunJary:
It still can't see clearly, I use the latest version (1.0.168)

dpdpbj:
anyone knows how to change the background color of sftp? I encountered the same issue.

SunJary:
I will fix it today later

PR #5320, opened 2021-12-28
fix: paper theme sftp color

SunJary:
fix Issues #4954 fix paper theme sftp background-color


dpdpbj:
Hi @SunJary, for the Tabby has been installed, do you know which file need to be changed to fix this issue? MacOS.

SunJary:
> Hi @SunJary, for the Tabby has been installed, do you know which file need to be changed to fix this issue? MacOS.

Sorry，I don't know that.

PR #5670, opened 2022-02-05
Fix app not quitting completely on Cmd+Q (Mac)

### LINKED ISSUE #4930: Tabby prevents Mac from getting shut down
Hi, I've noticed that having Tabby open prevents Mac from being shut down unless you quit the app manually.

When you shut down Mac, all Tabby windows along with other apps are closed, except the app keeps running in the background and prevents the shut down until the app is stopped manually. This happens regardless of whether you have any terminals open as long as the main app window is open.

I've noticed the same thing happens when you open the app switcher (<kbd>Cmd</kbd> + <kbd>Tab</kbd>), navigate to Tabby and press <kbd>Cmd</kbd> + <kbd>Q</kbd> to exit the app - it will close all the windows, but you need to press <kbd>Cmd</kbd> + <kbd>Q</kbd> again to kill the app, which seems to me like an incorrect behaviour - is there any particular reason it behaves that way?

mwz:
This change seems to fix #4930 for me locally.
This was initially added in https://github.com/Eugeny/tabby/commit/bcf09c59e35819981945a79d0db6a35333ff11b5 by @Eugeny - is there any specific reason for why you added the check to exclude Macs here?

closes #4930

### DIFF
--- a/app/lib/app.ts
+++ b/app/lib/app.ts
@@ -94,7 +94,7 @@ export class Application {
         }
 
         app.on('window-all-closed', () => {
-            if (this.quitRequested || process.platform !== 'darwin') {
+            if (this.quitRequested) {
                 app.quit()
             }
         })


Eugeny:
No clue why I have added it - IIRC it was there even before in `index.ts` and I just added `quitRequested`. Anyway, thanks a lot for the fix! @allcontributors please add @mwz for code.

Eugeny:
Right - it was there so that the app quits when closing all windows on other platforms. It couldn't have had any effect on macOS.

mwz:
Hi @Eugeny, how do you mean that it couldn't have had any effect on Mac - it fixed the issue for me when I tested it locally.
I can see that you reverted the change in 60a48ec2a5658b8c9497c520d7b91cc89d5ca3bf?

Eugeny:
Yes - the expression you removed was always `false` on macOS and hence the PR didn't do anything except prevent the app from exiting Tabby on Windows and Linux - sorry.

mwz:
> the expression you removed was always `false` on macOS

Yes that's correct - which means the app on Mac would stay running in the background when all windows were closed until the user explicitly quits it (which as a result prevents mac from shutting down automatically as described in #4930).

> and hence the PR didn't do anything except prevent the app from exiting Tabby on Windows and Linux

ok, sorry that's an oversight on my end - I didn't take other platforms into consideration there.
If that's the case, then could we maybe simplify this and make the behaviour consistent across all platforms by removing this logic entirely like so:

```ts
app.on('window-all-closed', () => {
   app.quit()
})
```

It seems that this `quitRequested` variable is ignored on Windows & Linux (due to `process.platform !== 'darwin'` being always `true`) and at the same time `process.platform !== 'darwin'` is preventing from the app quitting automatically when all windows are closed, so removing this should solve this issue and not break the existing behaviour on Windows and Linux. Apologies for breaking this.

@Eugeny, are you happy for me to raise a new PR with this change?

Eugeny:
I'm still not getting it.

on Windows/Linux, it's:
`if (this.quitRequested || true) {`

on macOS it's
 
`if (this.quitRequested || false) {`

which is already the same as 

`if (this.quitRequested) {`

The reason for the `quitRequested` check is to prevent the app from quitting just because all windows are closed - in line with other macOS apps' behaviour. 

It seems that listening for the `before-quit` event and then setting `quitRequested` could be a better idea, although I haven't tested whether that even really is sent when shutting down.

mwz:
I could never quite understand why that's a default Mac behaviour, but yeah sounds like your suggestion might be a better option to what I proposed. I don't have much electron experience, so thanks for bearing with me while I try to work this out.

I will give this a try and will get back to you on whether this works.

PR #5902, opened 2022-03-08
Fixing fluent not working as expected for Windows 11

microhobby:
Before the glasstron bump:

https://user-images.githubusercontent.com/2633321/157175135-6763f346-6cfb-420e-bd42-4b289f0b63b0.mp4

After the bump (also not disabling fluent while dragging):

https://user-images.githubusercontent.com/2633321/157175230-e6fce0cc-eb27-4793-8842-a52ae9b076d5.mp4

I need your help here @Eugeny because I don't know the story behind disabling dragging with fluent or if this was also fixed by bump of the glasstron.

### DIFF
--- a/app/lib/window.ts
+++ b/app/lib/window.ts
@@ -420,7 +420,7 @@ export class Window {
 
         let moveEndedTimeout: any = null
         const onBoundsChange = () => {
-            if (!this.lastVibrancy?.enabled || !this.disableVibrancyWhileDragging || !this.isFluentVibrancy) {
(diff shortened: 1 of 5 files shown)


Eugeny:
Great catch! I don't have the same bug (white background) on my Win11, but if the Glasstron upgrade fixes it, that's great. It has also fixed the laggy window drag behaviour that was the reason it's being disabled during drag. I'm going switch vibrancy over to stay enabled while dragging now, but keep it as an option under _Hacks_ :+1:

Eugeny:
@allcontributors please add @microhobby for code

Issue #5794, opened 2022-02-18
Fresh clean install on Windows 7 x64, fails to launch


dr-dba:
Fresh clean install on Windows 7 x64, fails to launch.
Here is the detailed output log i got:

Microsoft Windows [Version 6.1.7601]
Copyright (c) 2009 Microsoft Corporation.  All rights reserved.
C:\ProgFiles\tabby>tabby
[Glasstron/DWM] Trying to load the native module...
[Glasstron/DWM] Native module loaded
(node:9168) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. 
Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
(Use `Tabby --trace-deprecation ...` to show where the warning was created)
Administrator: C:\Windows\system32\cmd.exe[4384]: 
../../third_party/electron_node/src/node_buffer.cc:246: 
Assertion `val->IsArrayBufferView()' failed.
 1: 0000000141577536 node::Buffer::New+50294
 2: 0000000141577279 node::Buffer::New+49593
 3: 0000000141569E82 node::Buffer::Data+178
 4: 000007FEF501738D
 5: 0000000142B7D58D v8::MicrotasksScope::GetCurrentDepth+98301
 6: 0000036E07F0B33C
[1128:0218/095801.874:ERROR:angle_platform_impl.cc(44)] 
SwapChain11.cpp:941 (present): 
Present failed: the D3D11 device was removed, HRESULT: 0x887A0020
[1128:0218/095801.875:ERROR:gl_surface_egl.cc(808)]
EGL Driver message (Critical) eglPostSubBufferNV: Context lost.
[1128:0218/095801.875:ERROR:shared_context_state.cc(834)] 
SharedContextState context lost via ARB/EXT_robustness. 
Reset status = GL_UNKNOWN_CONTEXT_RESET_KHR
[1128:0218/095801.877:ERROR:gpu_service_impl.cc(964)] 
Exiting GPU process because some drivers can't recover from errors. 
GPU process will restart shortly.

Is Windows 7 x64 supported?
Can it be cured?
Thank you

Issue #5528, opened 2022-01-18
Crashed on startup with Win7 after 1.0.168


aisatis:
Crashed when startup.It will show startup animate,but disapper soon.
It's ok under 1.0.168,but fail under 1.0.169 and 1.0.170.
My OS is Win7 Professional,lang is Zh-CN.
I've tried to remove Folder "C:\Users\<my_account>\AppData\Roaming\tabby" but it's useless,seems it's not the fault of plugins.

==================
running result in CMD

[Glasstron/DWM] Trying to load the native module...
[Glasstron/DWM] Native module loaded
(node:14024) [DEP0005] DeprecationWarning: Buffer() is deprecated due to securit
y and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or
Buffer.from() methods instead.
(Use `Tabby --trace-deprecation ...` to show where the warning was created)
(node:14024) DeprecationWarning: matchedHotkey is deprecated, use hotkey$
绠＄悊鍛? C:\Windows\System32\cmd.exe[17480]: ../../third_party/electron_node/sr
c/node_buffer.cc:246: Assertion `val->IsArrayBufferView()' failed.
 1: 0000000140964886 node::Buffer::New+49750
 2: 00000001409645C9 node::Buffer::New+49049
 3: 0000000140957407 node::Buffer::Data+167
 4: 000007FEC552738D
 5: 0000000141B0FB23 v8::EmbedderHeapTracer::IncreaseAllocatedSize+71491
 6: 00000610000C173C

aisatis:
seems related to Win7.
It works fine on Win10, but also crashed on my another PC with Win7.
Is it related to Electron's version?

wind0ta:
I also have the same experience, windows 7, version 170.

xavierfoucrier:
Hi @Eugeny,

And first of all, thanks for this awesome terminal software.
I can confirm the issue on Windows 7 x64 on (and after) Tabby `1.0.168`.

Let me know if I can help on this :wink:

xavierfoucrier:
Still an issue with the latest Tabby `1.0.172`.

Issue #5465, opened 2022-01-13
Crashing when opening tabby in windows 7


shenweiyan:
**Describe the problem**:
I try [tabby-1.0.169-setup.exe](https://github.com/Eugeny/tabby/releases/download/v1.0.169/tabby-1.0.169-setup.exe) and [tabby-1.0.169-portable.zip](https://github.com/Eugeny/tabby/releases/download/v1.0.169/tabby-1.0.169-portable.zip) on my Windows 7, 64bit, both of them crashing when opening Tabby.exe, does tabby still support win7? 

Please, can anyone help me with this problem？

**To Reproduce**:
[Steps to reproduce the behavior]

shenweiyan:
> **Describe the problem**: I try [tabby-1.0.169-setup.exe](https://github.com/Eugeny/tabby/releases/download/v1.0.169/tabby-1.0.169-setup.exe) and [tabby-1.0.169-portable.zip](https://github.com/Eugeny/tabby/releases/download/v1.0.169/tabby-1.0.169-portable.zip) on my Windows 7, 64bit, both of them crashing when opening Tabby.exe, does tabby still support win7?
> 
> Please, can anyone help me with this problem？
> 
> **To Reproduce**: [Steps to reproduce the behavior]

I try 1.0.168, it works.

vsravuri:
I'm also facing this issue, 1.0.168 works for me on Windows 2012 R2 but crashing with 1.0.169 and 1.0.170

boonkerz:
any details in the event log ?

vsravuri:
@boonkerz 

I don't see any errors in Windows -> Event Viewer
Tabby log file is empty.

After launching Terminus.

**1.0.168**
<img width="737" alt="Screen Shot 2022-01-16 at 2 55 12 PM" src="https://user-images.githubusercontent.com/22808594/149681730-a3af5ac9-2bb1-47a8-baa1-1cade1590d17.png">


**1.0.170**
<img width="736" alt="Screen Shot 2022-01-16 at 2 54 50 PM" src="https://user-images.githubusercontent.com/22808594/149681743-ab2e385f-afcb-424f-9aec-e077e3ad42d1.png">

Could it be possible the user profile path not found in 1.0.169+ is causing the crash ?

nsusn:
![6](https://user-images.githubusercontent.com/33510340/149701286-6926e609-1d76-48ec-93ad-03f8bb3f874c.png)
[...]

PR #6027, opened 2022-03-26
fix crash on Windows 7

### LINKED ISSUE #5339: Alpha 169 crashes on startup on Windows 7
**Describe the problem**:
Alpha 169 crashes on startup on Windows 7 machine. "data\log.txt" is empty. Version 168 works normally.

**To Reproduce**:
Crashes every time on app launch.

ysc3839:
After debugging I found the crash on win7 is caused by `DwmEnableBlurBehindWindow` in `window.ts`.
`window.ts` pass `this.window` to `windows-blurbehind`, then `windows-blurbehind` pass this object to `node::Buffer::Data`:
https://github.com/Eugeny/windows-blurbehind/blob/30a3dd4f357493d796c1e4770f478bdf6840d4af/src/blurbehind.cc#L7
`node::Buffer::Data` found it's not a `ArrayBufferView` and crash:
https://github.com/nodejs/node/blob/v16.13.0/src/node_buffer.cc#L246

Tabby 1.0.167 uses electron 13.5.1, I don't know why it works. According to electron [docs](https://www.electronjs.org/zh/docs/latest/api/browser-window#wingetnativewindowhandle), here should use `win.getNativeWindowHandle()` to get `Buffer` of window handle.

Fixes #5339, #5465, #5528, #5794

### DIFF
--- a/app/lib/window.ts
+++ b/app/lib/window.ts
@@ -182,7 +182,7 @@ export class Window {
                     console.error('Failed to set window blur', error)
                 }
             } else {
-                DwmEnableBlurBehindWindow(this.window, enabled)
+                DwmEnableBlurBehindWindow(this.window.getNativeWindowHandle(), enabled)
             }
         } else if (process.platform === 'linux') {
             this.window.setBackgroundColor(enabled ? '#00000000' : '#131d27')


Eugeny:
Great find - thanks for fixing! :+1: 

@allcontributors please add @ysc3839 for code

PR #6046, opened 2022-03-29
Preserve profile groups collapse state.

artu-ole:
Small quality of life improvement to remember which profile groups were collapsed.

Motivation:
I have many different connection profiles in many different groups that need to be modified/copied quite often hence starting at a partially collapsed view allows me to pickup where I stopped last time and ignore less relevant profiles easier.


Eugeny:
Thanks! This is looking great, but should use `LocalStorage` instead of the config file. Storing small things like these in the config will mean that e.g. a config sync will get triggered every time you open/close a group.

artu-ole:
Yep, good point, I also didn't like how refresh would run every time it toggled, but was trying to use same tools already used in the component. I'll change this to use localStorage

artu-ole:
Fixed the storage, let me know if there's anything else I can improve.

Eugeny:
Looking great now - thanks a lot! 

@allcontributors please add @artu-ole for code

PR #6140, opened 2022-04-12
Allows the user to resize panes using hotkeys

ChristianBingman:
Changes:
- Adds 4 hotkeys for resizing panes horizontally and vertically
- Updates the config files (default not set)
- Updates the window settings to allow the user to modify the pane
  resize increments


Eugeny:
Working nicely, thanks! I've renamed the setting and put it under the `terminal` key, and removed the `ratio` rebalance code (it was broken and ratios are rebalanced in `.layout()` anyway) 🙌 

@allcontributors please add @ChristianBingman for code

PR #6175, opened 2022-04-15
Fix small typo

2ykwang:
corrected the wrong words.

ref:
https://en.wikipedia.org/wiki/Korean_language


Eugeny:
Sorry!! 😭 I took it from https://en.wikipedia.org/wiki/List_of_language_names

2ykwang:
Thank you @Eugeny Tabby is very useful. 👍

PR #6199, opened 2022-04-19
Update ru-RU.po

Zalexanninev15:


Eugeny:
См. https://github.com/Eugeny/tabby/blob/master/locale/STOP.txt

PR #6207, opened 2022-04-20
Update zh-CN.po

Remalloc:
Add Chinese translation.


Eugeny:

PR #6218, opened 2022-04-20
fix import lowcase host from ssh/config

Ox0400:
support import lowercase host field. from issue: https://github.com/Eugeny/tabby/issues/6219
```
Host upperhostfield
    ...

host lowerhostfield
   ....
```


Eugeny:
Thanks! @allcontributors please add @Ox0400 for code

PR #6232, opened 2022-04-21
Update ja-JP.po

eltociear:
Github -> GitHub


Eugeny:

Issue #6205, opened 2022-04-19
Handle local env variable in SSH profile configuration


woodmeal:
**Is your feature request related to a problem? Please describe.**
I've to deal with several SSH connections with the same user and I'd like to be able to just have to write $USER or something like this when I'm adding a new SSH profile.

![screenshot cleaned](https://user-images.githubusercontent.com/104011197/164017641-3b7f277a-65d0-42a8-9294-ec01558ff424.png)

**Describe the solution you'd like**
Adding an `if` statement to check if the first char is a `$` and replace it by the env variable of the input would be the easiest solution though I'm not sure if it's possible to read env variables(that are not in a .env file) from an electron app.

woodmeal:
I also thought about a checkbox under username to use the current user

woodmeal:
if anyone can guide me through the tabby's code to find where this window is located, I'll be glad to add this feature through PR.

PR #6242, opened 2022-04-22
Adding env handlement for ssh profile sessions

woodmeal:
As described in the issue #6205 , I added the possibility to handle environment variables just by adding a '$' in the SSH profile configuration.

My co-workers and I are already using it, even if it could be improved in the frontend part

If any changes are needed for this PR I'll be glad to make them :)


Eugeny:
Thanks! @allcontributors please add @woodmeal for code

matthiscock:
Can this also be used in the Private keys path?
c:\users\%USERNAME%\.ssh\id_rsa for example 

(right now doesnt work)

PR #6345, opened 2022-05-03
Fix a translate mistake

calvinweb:


Eugeny:

Eugeny:
Please disregard my previous message :D - thank you for the contribution!

PR #6465, opened 2022-05-19
Pane navigation improvements

attet:
Hi there, thanks for creating such a great terminal!

In case you think these might be useful for others as well, I added hotkeys for navigating to specific panes and a 'focus follows mouse' -option. Both unassigned/off by default but they work very well for me.

### DIFF
--- a/tabby-core/src/components/splitTab.component.ts
+++ b/tabby-core/src/components/splitTab.component.ts
@@ -311,6 +311,33 @@ export class SplitTabComponent extends BaseTabComponent implements AfterViewInit
                 case 'pane-nav-next':
                     this.navigateLinear(1)
                     break
+                case 'pane-nav-1':
+                    this.navigateSpecific(0)
+                    break
+                case 'pane-nav-2':
+                    this.navigateSpecific(1)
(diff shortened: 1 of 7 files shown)


Eugeny:
Wow, thanks for the PR - looks great! My only nitpick is that the hotkey definitions should use ICU plurals so that translators don't have to translate them 10 times :) - like here: https://github.com/Eugeny/tabby/blob/master/tabby-core/src/hotkeys.ts#L69

attet:
> Wow, thanks for the PR - looks great! My only nitpick is that the hotkey definitions should use ICU plurals so that translators don't have to translate them 10 times :) - like here: https://github.com/Eugeny/tabby/blob/master/tabby-core/src/hotkeys.ts#L69
> 
> >  GitHub**[tabby/hotkeys.ts at master · Eugeny/tabby](https://github.com/Eugeny/tabby)**A terminal for a more modern age. Contribute to Eugeny/tabby development by creating an account on GitHub.

Thanks, fixed.

Eugeny:
@allcontributors please add @attet for code

PR #6493, opened 2022-05-25
Hide description in selector modal if it's the same as the title

bdr99:
I have a lot of SSH profiles where the name of the profile is the same as the SSH hostname, which clutters the UI. This PR cleans it up a bit by hiding the description (hostname) if it is the same as the title (name of profile).


Eugeny:
Much appreciated :+1:

Issue #6653, opened 2022-06-22
Add option(s) in tab context menu to kill running process and/or restart the tab


Totalus:
**Is your feature request related to a problem? Please describe.**

I sometimes run a command that I am unable to stop, even with `Ctrl+C` multiple times, it just keeps running. The only way I can stop it is to close the tab.

**Describe the solution you'd like**

I would like to be able to force kill the process running in a tab without having to close my tab and open a new one.

**Describe alternatives you've considered**

Currently I use the _Duplicate_ option to duplicate my tab, which copies most of my context (title, color, path, etc.) to a new tab, and then I close the original tab.

**Additional details**

This could be easily implemented with an additional option in the tab's context menu. We could have one or possibly event both of the following:
- `Restart` : Would restart the tab (similar to duplicating + closing original tab). This would keep most of the context (color, title, path), but would reset the environment variables.
- `Kill process` : Would only force kill the running process. This would keep the context and the environment variables intact.

Qiming-Liu:
For PR https://github.com/Eugeny/tabby/pull/6709

![image](https://user-images.githubusercontent.com/68600416/176991936-1e2178b0-20a9-4d22-b958-f8a22b5f0aef.png)

Any advice or better solution feel free to let me know

Totalus:
Looks good to me 👍

Qiming-Liu:
> Looks good to me 👍

Thanks!
As here https://github.com/Eugeny/tabby/pull/6709#issuecomment-1175339271
This will become a configurable hotkey function.

Qiming-Liu:
This issue can be closed.

Remonli:
Can not find such restart tab in context menu.

env:  1.0.182 x64 portable on Win10 x64 2021 LTSC

<img width="357" alt="1659273375413" src="https://user-images.githubusercontent.com/9067053/182028241-3f4952f6-0aa4-4798-a0a5-4379aeaf0926.png">

Totalus:
This feature was added as a hotkey instead of a context menu item ([reference](https://github.com/Eugeny/tabby/pull/6709#issuecomment-1175771106)). Not sure in which version it is released.

Qiming-Liu:
> This feature was added as a hotkey instead of a context menu item ([reference](https://github.com/Eugeny/tabby/pull/6709#issuecomment-1175771106)). Not sure in which version it is released.

Thank you. That is true. It's in Alpha 182.
[...]

PR #6709, opened 2022-07-02
feat: tab context menu 'Restart'

Qiming-Liu:
For issue https://github.com/Eugeny/tabby/issues/6653#issue-1280295196

image

### DIFF
--- a/tabby-core/src/components/appRoot.component.ts
+++ b/tabby-core/src/components/appRoot.component.ts
@@ -112,6 +112,10 @@ export class AppRootComponent {
                 if (hotkey === 'duplicate-tab') {
                     this.app.duplicateTab(this.app.activeTab)
                 }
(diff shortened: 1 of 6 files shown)


Eugeny:
For this option, I would prefer a hotkey instead of a menu entry - as it's rarely necessary for local shells, and remote tabs already have their own dedicated restart button.

Qiming-Liu:
> For this option, I would prefer a hotkey instead of a menu entry - as it's rarely necessary for local shells, and remote tabs already have their own dedicated restart button.

Agree, I’ll try to make a hotkey instead of this tomorrow. Thanks for your reply!

Qiming-Liu:
image


Tested on windows works fine.
No default hotkey value.

Please let me know if this is ok.

Eugeny:
This is great, thanks!

PR #6877, opened 2022-07-30
adding a securecrt to tabby converter

mc36:


Eugeny:
Appreciate the PR, but I'd prefer to have it either as a profile importer similarly how it's done for OpenSSH (https://github.com/Eugeny/tabby/blob/master/tabby-electron/src/sshImporters.ts) or as a link to your own repo.

PR #6885, opened 2022-08-01
Save downloaded config (fix #6694)

### LINKED ISSUE #6694: Failed to synchronize configuration
Version: 1.0.181
Platform: Windows x64 10.0.22000
Plugins: sync-config
Frontend: xterm

Windows 11 A，Configuration uploaded
image

Windows 11 B，After the configuration is downloaded, it is still not configured
image
image

giejqf:
The downloaded config may fail to be saved because of a race condition in writeConfigDataFromSync().
Inside this function, the first line `await this.platform.saveConfig(yaml.dump(data)) `performs the save action in an async way (ipcRenderer.send / ipcMain.on). It is possible that the new config is not saved before invoking `await this.config.load()`, causing it to load the old config and then override the config file in `await this.config.save()`.

This commit replaces (ipcRenderer.send / ipcMain.on) with (ipcRenderer.invoke / ipcMain.handle) to ensure that the new config is saved first.

### DIFF
--- a/app/lib/app.ts
+++ b/app/lib/app.ts
@@ -31,7 +31,7 @@ export class Application {
         this.useBuiltinGraphics()
         this.ptyManager.init(this)
 
-        ipcMain.on('app:save-config', async (event, config) => {
+        ipcMain.handle('app:save-config', async (event, config) => {
             await saveConfig(config)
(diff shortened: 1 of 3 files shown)


Eugeny:
Great find, much appreciated! @allcontributors please add @giejqf for code

Issue #7110, opened 2022-09-09
SSH profile import supports ssh_config keyword 'include'


llouislu:
**Is your feature request related to a problem? Please describe.**
I have SSH profiles in `~/.ssh/config`  where they are spread out in several sub-config files using `Include` from [ssh_config(5)](https://man7.org/linux/man-pages/man5/ssh_config.5.html).
Here is a minimal sample.
In `~/.ssh/config`
```
Include ~/.ssh/test_nodes

Host node1
...
...
...
```
In `~/.ssh/test_nodes`
```
Host testnode1
...
...
...
```
Tabby only loads `node1` in profile-selector. This feature allows all other profiles loaded from included configs in the main config, such as `testnode1`.

**Describe the solution you'd like**
`OpenSSHImporter` recursively finds all included profiles from `~/.ssh/config` so that all the profiles from included config files are shown up in the profile-selector.

**Describe alternatives you've considered**
Manually add profiles in.

**Additional context**
N\A

PR #7111, opened 2022-09-09
WIP:  added support to ssh profiles from ssh_config's `include` #7110

llouislu:
This loads ssh profiles included in extra config files from `~/.ssh/config` https://github.com/Eugeny/tabby/issues/7110

Work in progress now.


Eugeny:
Already handled in #10105

Issue #6126, opened 2022-04-09
WinSCP intergration not working when using SSH keys


kyuuaria:
**Describe the problem**:
Launching WinSCP from the context menu doesn't work when using SSH keys [as seen here](https://i.imgur.com/KnOwq1H.png)

**To Reproduce**:
Set up an SSH connection profile with an SSH key as the authentication method connect to it and try to launch WinSCP from the context menu

kyuuaria:
The private key probably doesn't get passed right https://winscp.net/eng/docs/commandline#parameters

kyuuaria:
I found using pageant to be a good workaround, but it still sucks

Issue #5106, opened 2021-12-10
Can`t open winscp with publickey


WhiteBite:
When i try "open in winscp" i get this
![image](https://user-images.githubusercontent.com/19576772/145588064-8884c917-23c1-4a15-a833-d9f2082523d4.png)
Translate error
```
The authentication protocol (see the connection log for details:
I am using the username "root".

Authentication failed.
```

Issue #3480, opened 2021-02-21
SSH key transfer to WinSCP


dargow:
The function of transferring the private key is required when using it as authorization to the program WinSCP. 
At the moment WinSCP ask password.
Similar functionality is implemented in the terminal Kitty

kyuuaria:
apparently not and I guess nothing will change in this respect

Issue #6319, opened 2022-04-29
Add full Sftp client support


saudiqbal:
It would be nice to have two columns, Remote and Local and something like WinSCP. Drag and drop files/folders from remote/local and other common options.

bedbug6404:
That would be great! Something like Mobaxterm would be neat and handy!

PR #7143, opened 2022-09-19
Support private key forwarding to winscp

artu-ole:
Fixes https://github.com/Eugeny/tabby/issues/5106
Fixes https://github.com/Eugeny/tabby/issues/3480
Fixes https://github.com/Eugeny/tabby/issues/6126

Naive ssh key transfer via temporary file converting it to ppk format that is the only one supported in cli arguments in winscp. https://winscp.net/eng/docs/public_key#private
There are couple of additional dependancies in the code.
First one, tmp-promise, already used in some other modules, like tabby-electron. Used for the temporary file creation as winscp does not accept private key directly from stdin
Second one, node-forge, merely used to convert the private key into putty format and is something that would not be needed in an ideal world. Unfortunatelly, sshpk putty private keys support(https://github.com/TritonDataCenter/node-sshpk/pull/76) did not come with proper putty files writer. We'd need to either make a pull request to that upstream project or to Eugene's fork of it.

And finally, even though current sftp functionality is sufficient for 95% of work, making a more sophisticated sftp client(https://github.com/Eugeny/tabby/issues/6319) would be the way to go in the future for a one stop shop experience.

P.S. also, ugly timeout is there not to dispose of a temporary file beforehand and because we don't get a child process reference to hook into.


Eugeny:
Thanks! I'm on vacation until net week and will check it out then.

artu-ole:
Sure, no rush, In the end this might be more of a proof of concept rather than final implementation due to aforementioned shortcomings.
Getting rid of timeout is fairly simple, should be good enough if exec returned the result of execFile, something along the lines:
```ts
await this.platform.exec(path, args)
tmpFile?.cleanup()
```
Though that would make tabby wait to remove the file until winSCP is closed(or tabby itself is closed as tmp-promise library states that it will clean files on process exit by itself)

There might be a deeper issue lurking. It does not feel entirely safe saving user's private key into a temporary directory rather than some secure application storage. Though I'm not too well versed with electron to suggest a better solution right away.

Eugeny:
Made `exec` async. It's very unfortunate that WinSCP doesn't accept private keys through environment vars which would be both convenient and perfectly safe. A workaround could be to use WinSCP's agent auth by spinning up a temporary SSH agent, but this will conflict with an already running Pageant since there's no way to run multiple agents on Windows.

Eugeny:
Tested and it LGTM - let me know if the changes look ok to you and I'll merge :v:

artu-ole:
Yep, LGTM, works like a charm

Issue #5931, opened 2022-03-13
Windows - Terminal bell sound not working


Figoware:
**Describe the problem**:
No audible sound even if terminal option set to Aubible.
Nohting in Windows sound mixer (if it could help) 
Tested on 2 Windows 10, Tabby alpha 173

**To Reproduce**:
Settings, Terminal, Sound, set to Aubible.
Go to a terminal, use tab for completion: no sound

Change setting to "Visual": window shakes

Figoware:
Still not working with 1.0.181.

Console says:
`AppData\Local\Programs\Tabby\resources\builtin-plugins\tabby-terminal\dist\index.js:16086 Uncaught (in promise) DOMException: The element has no supported sources.`

...triggered by 
```js
if (this.config.store.terminal.bell === 'audible') {
                this.bellPlayer.play();
}
```

PR #7368, opened 2022-10-26
Fix for audio bell.ogg #5931

zhoro:
Small fix for Terminal bell


Eugeny:
Thank you! @all-contributors please add @zhoro for code

PR #7398, opened 2022-10-31
use winSCP.com utility to convert private key to ppk to launch winSCP

### LINKED ISSUE #7204: Private key forwarding to winscp isn't working
**Describe the problem**:
The private key forwarding to winscp added in #7143 doesn't work for me. Clicking `Launch WinSCP` simply doesn't do anything. There is also nothing in the logs. It should be noted that I got both my private key and the corresponding passphrase in the vault. I also made sure to add winscp to path because I thought it might not be able to execute the command.

artu-ole:
Fixes #7204

Using [winSCP.com](https://winscp.net/eng/docs/executables#winscp.com) utility that comes with winSCP to convert private key to ppk for winSCP connection for it's support of many different types of key types(RSA, DSA, ECDSA, Ed25519).
As an additional improvement it could show error message in case something goes wrong in a popup/modal rather than just in the console, but I didn't make that more involved change.

Tested with rsa, ecdsa and ed25519 keys in different formats, both password protected and passwordless(sshpk takes care of unencrypting the key beforehand).
Drop node-forge dependency as redundant.

### DIFF
--- a/tabby-ssh/package.json
+++ b/tabby-ssh/package.json
@@ -29,7 +29,6 @@
   },
   "dependencies": {
     "@luminati-io/socksv5": "^0.0.7",
-    "node-forge": "^1.3.1",
     "run-script-os": "^1.1.3",
     "tmp-promise": "^3.0.3"
   },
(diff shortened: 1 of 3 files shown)


Eugeny:
LGTM :+1:

Issue #2704, opened 2020-06-15
Check on Hotkeys


brc-dd:
While setting a custom hotkey, Terminus doesn't check if the same hotkey is there for any other operation. There is an error in default configuration that for `new-tab` and `reopen-tab` the hotkey is same `Ctrl-Shift-T`. It should rather be `Ctrl-T` and `Ctrl-Shift-T` respectively. Such errors can unknowingly occur on the part of the users also. A simple loop to check if hotkey exists on the configuration can solve this issue. You guys should work on this issue (surely this is not a top priority 🙂 ).

PR #7667, opened 2022-12-18
Mark duplicate hotkeys

BenjaminBrandmeier:
Feature to mark duplicate hotkeys. Solves https://github.com/Eugeny/tabby/issues/2704.

hotkey_duplicates

### DIFF
--- a/tabby-settings/src/components/hotkeySettingsTab.component.ts
+++ b/tabby-settings/src/components/hotkeySettingsTab.component.ts
@@ -3,11 +3,12 @@ import { marker as _ } from '@biesbjerg/ngx-translate-extract-marker'
 import { Component, NgZone } from '@angular/core'
 import {
     ConfigService,
+    Hotkey,
     HotkeyDescription,
     HotkeysService,
     HostAppService,
 } from 'tabby-core'
-import { Hotkey } from 'tabby-core/src/api/hotkeyProvider'
+import deepEqual from 'deep-equal'
 
 _('Search hotkeys')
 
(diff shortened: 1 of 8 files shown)


tabby-settings/src/components/hotkeySettingsTab.component.ts
```diff
@@ -30,28 +31,44 @@ export class HotkeySettingsTabComponent {
...
     }
+
+    private detectDuplicates (strokes: string[] | string): Hotkey {
+        const allHotkeys = Object
+            .values(this.config.store.hotkeys)
+            .filter((value: unknown) => Array.isArray(value))
+            .flat()
+
+        const isDuplicate = allHotkeys
+            .filter(hotkey => JSON.stringify(hotkey) === JSON.stringify(strokes))
```
Eugeny, line 69: I'd prefer `deepEquals` here since JSON won't guarantee key order if hotkeys would contain dicts some day.

tabby-settings/src/components/multiHotkeyInput.component.ts
```diff
@@ -1,6 +1,7 @@
 import { Component, Input, Output, EventEmitter, ChangeDetectionStrategy } from '@angular/core'
 import { NgbModal } from '@ng-bootstrap/ng-bootstrap'
 import { HotkeyInputModalComponent } from './hotkeyInputModal.component'
+import { Hotkey } from 'tabby-core/src/api/hotkeyProvider'
```
Eugeny, line 4: `Hotkey` should be barrel-exported through `tabby-core/src/index.ts`, same as everything else

Eugeny (changes requested):
Otherwise LGTM 👍

Eugeny:
Great work!

Eugeny:
Unfortunately it's now making filtering unusable - do you think you could make it faster than O(N^2)? E.g. stick all keystrokes into a Set and look up from it.

Screenshot 2023-01-06 at 16 44 41

BenjaminBrandmeier:
Absolutely! Will create a performance fix PR within the next minutes.

PR #7757, opened 2023-01-06
Improve hotkey filtering performance

BenjaminBrandmeier:
This is related to https://github.com/Eugeny/tabby/pull/7667 which introduced a performance regression when filtering hotkeys.

Implementation details:

- duplicate detection happens once on start and each time hotkeys are changed
- hotkey comparison happens on strings. (deep-equal is slow when it comes to checking array equality)


Eugeny:
Thanks!

Issue #7254, opened 2022-10-07
Sync Host Box


WinterFe:
**Describe the problem**:
Settings -> Config Sync -> Sync host
Typing in the sync host box ( Win 11 ) is nearly impossible, it just deletes previous characters and is buggy as hell.

**To Reproduce**:
Just type. Lol.

Related Gif:
https://i.lunardev.group/VCJqMSL/direct

PR #7791, opened 2023-01-15
fix Eugeny/tabby#7254 - config sync replace ngModelChange by keydown enter

Clem-Fern:
Hi,

First of all, thank's for the work on Tabby. I really love it !

I encountered the same problem as the one describe in the issue #7254.
This fix seems acceptable to me since we don't need to save the config and test the connection on each inserted or removed character in the host and token input field. But i possibly missed something so feel free to ask me if any change needed :)


Eugeny:
Thank you for the contribution and sorry for the delay! @all-contributors please add @Clem-Fern for code

PR #7844, opened 2023-01-28
feat: publish typings for tabby-ssh

elizabeth-dev:
I'm developing a small plugin for Tabby (one that fetches instance data from AWS in order to make it easier to start SSH connections), and I found out that, even though the `tabby-ssh` plugin is written in Typescript and has typings declared in the `package.json`, those typings are not included when publishing the package. This makes development of plugins that consume the `tabby-ssh` plugin API a bit harder.

This change includes the `typings` folder in the `tabby-ssh` package, so the Typescript type definition files are bundled in it.

To be honest this probably should be done for all the plugins in this repo, but I've only checked `tabby-ssh`. I'm open to include those other plugins in this PR if you consider that useful.

### DIFF
--- a/tabby-ssh/package.json
+++ b/tabby-ssh/package.json
@@ -16,7 +16,8 @@
   },
   "files": [
     "dist",
-    "util/pagent.exe"
+    "util/pagent.exe",
+    "typings"
   ],
   "author": "Eugene Pankov",
   "license": "MIT",


elizabeth-dev:
Additionally to the PR, I'm having trouble having my provider injected as a SSH Importer, even though I'm extending `SSHProfileImporter`. Is creating custom SSH Importers not supported? If it's not, would you be open to supporting it?

Eugeny:
Thank you for the contribution! I've published the new version with typings.

Make sure you're exposing your importer in your plugin's module description, like it's done here: https://github.com/Eugeny/tabby/blob/master/tabby-electron/src/index.ts#L35-L36

elizabeth-dev:
I actually used that as a reference, so I did everything the same way... I must be missing something

The (WIP) code for what I'm developing is here:
https://github.com/elizabeth-dev/tabby-aws/blob/main/src/index.ts#L18

(In any case I know this isn't the place to ask for help, so sorry for that, I guess I'll just keep trying)

Eugeny:
Found it - the SSH importers list was finalized too early, before external plugins got loaded - it's fixed now in the nightly (make sure you update your tabby-ssh dependency as well!)

I'm actually seeing my AWS hosts with the plugin now

elizabeth-dev:
Yayyyyy, thank you 😊

PR #7857, opened 2023-01-30
Fix various issues

siccous:
Hello Eugeny,

today I've tried to build updated version of Tabby and hit few bugs (using a Windows machine):

1. [can't build on Windows because of this bug in Electron](https://github.com/electron-userland/electron-builder/issues/7175)

bumbed Electron to 21.3.1

2. issue in SSH2 library

[your PR](https://github.com/mscdex/ssh2/pull/1200), but there was missing one last commit from your PR which fixed it for agent authentication too

3. small issue in substituteEnv function

if the ENV value is an integer (it happened for me during launch of Powershell profile), then there is no replace function, I am not a JS programmer, so maybe this is not the proper way how to fix it :)


Eugeny:
Looking great, thank you! :+1:

Eugeny:
@all-contributors please add @siccous for code

PR #7858, opened 2023-01-30
feat: add button to delete remote config sync

Clem-Fern:
Hi @Eugeny,

Hope you're doing well !

I think it would be great to be able to delete synch configuration directly from tabby client.
This PR simply add a button to delete the config from the API. I deliberately did not clean config on the client side when deleting as it was more appropriate to do like that in my mind. Feel free ta ask me if change are needed :)

image


Eugeny:
Wow, nice work! :+1:

PR #7860, opened 2023-01-30
fix: prevent backspace in profile selection

Clem-Fern:
Hello hello,

It's me again :D

I thought it would be a good idea to prevent the backspace event in the profile selection field when the filter is empty and the selected profile editable. Without preventing it, the last character of the address is deleted.

image

Backspace result in 1.0.188:
image

Backspace result with this commit:
image


Eugeny:
LGTM 🚀

PR #7871, opened 2023-02-01
 fix #4684 - removing profile's hotkeys on profile delete

### LINKED ISSUE #4684: (Bug) Deleting a profile via the "Profiles" tab doesn't remove it's hotkeys from config.yaml
Version: 1.0.158
Platform: Windows x64 10.0.19043

After deleting a profile via "Profiles" tab, it's hotkeys for "New tab" still remain in the config.yaml file. This can sometimes lead to binds not working.

Create a profile and assign a shortcut for "New tab: <profile>". Delete this profile and create a new one.  To the new profile, assign the same key combination for "New tab: <profile>".

The old profile is removed from the "profiles" list, but not from "hotkeys.profile".

This is what that looks like:
```
profile:
    local:git-bash: []
    local:custom:test-1:e574684e-0982-4fcc-9cc8-8665307c63e2:
      - - Ctrl-Shift-]
    local:custom:test-2:3dc1c968-8610-4686-a1a3-f2310078a1b6:
      - - Ctrl-Shift-]
```
Where "test-1" is the profile that was created and then deleted, and "test-2" the profile that was created afterwards.
As a result, the key combination doesn't do anything when pressed.

I understand it can be manually fixed by deleting it from the config file, but I wanted to point it out anyways.

Clem-Fern:
Hi @Eugeny,

This PR aims to fix the issue Eugeny/tabby#4684 by removing profile's hotkeys if some are present when the profile is deleted.
I moved the function `getProfileHotkeyName` in the abstract class HotkeyProvider. In that way, the function can be used through the api.

If I understood properly, using `delete` operator to dynamically remove keys from an object is not recommended. So feel free to ask if changes needed or if you want me to find a better way to delete hotkeys.

### DIFF
--- a/tabby-core/src/api/hotkeyProvider.ts
+++ b/tabby-core/src/api/hotkeyProvider.ts
@@ -1,3 +1,5 @@
+import { PartialProfile, Profile } from './profileProvider'
+
 export interface HotkeyDescription {
     id: string
     name: string
@@ -14,4 +16,8 @@ export interface Hotkey {
  */
 export abstract class HotkeyProvider {
     abstract provide (): Promise<HotkeyDescription[]>
+
+    static getProfileHotkeyName (profile: PartialProfile<Profile>): string {
+        return (profile.id ?? profile.name).replace(/\./g, '-')
+    }
 }
--- a/tabby-core/src/hotkeys.ts
+++ b/tabby-core/src/hotkeys.ts
@@ -2,7 +2,6 @@ import { Injectable } from '@angular/core'
 import { TranslateService } from '@ngx-translate/core'
 import { ProfilesService } from './services/profiles.service'
 import { HotkeyDescription, HotkeyProvider } from './api/hotkeyProvider'
-import { PartialProfile, Profile } from './api'
 
 /** @hidden */
 @Injectable()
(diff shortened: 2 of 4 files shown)


Eugeny:
Have you checked that it works? `config.store` is a proxy object with dynamic properties (defined in https://github.com/Eugeny/tabby/blob/master/tabby-core/src/services/config.service.ts#L32-L59) and `delete` likely won't delete the underlying value.  

I'd also rather export `AppHotkeyProvider` from `tabby-core` than add unrelated functionality to the base class

Clem-Fern:
> Have you checked that it works? `config.store` is a proxy object with dynamic properties (defined in https://github.com/Eugeny/tabby/blob/master/tabby-core/src/services/config.service.ts#L32-L59) and `delete` likely won't delete the underlying value.
> 
> I'd also rather export `AppHotkeyProvider` from `tabby-core` than add unrelated functionality to the base class
> 
> >  GitHub**[tabby/config.service.ts at master · Eugeny/tabby](https://github.com/Eugeny/tabby)**A terminal for a more modern age. Contribute to Eugeny/tabby development by creating an account on GitHub.

Yup, it works. It seems that using `delete` call the `__setValue`method defined into the proxy to set the `profile` object without the deleted property. But i can see now why it is ambiguous.
Does deleting the property as below could be more appropriate ?
```
if (this.config.store.hotkeys.profile.hasOwnProperty(profileHotkeyName)) {
       const profileHotkeys = deepClone(this.config.store.hotkeys.profile)
       // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
       delete profileHotkeys[profileHotkeyName]
       this.config.store.hotkeys.profile = profileHotkeys
}
```

Eugeny:
Got it, thanks! If you could just move `getProfileHotkeyName` then we're good to go.

Clem-Fern:
Yeah sure, no problem.
I'm opening a new pull request with the adequate commits :)

PR #7884, opened 2023-02-04
fix #4684 - removing profile's hotkeys on profile delete

### LINKED ISSUE #4684: (Bug) Deleting a profile via the "Profiles" tab doesn't remove it's hotkeys from config.yaml

Clem-Fern:
As discuss in #7871, fix #4684

### DIFF
--- a/tabby-core/src/index.ts
+++ b/tabby-core/src/index.ts
@@ -258,6 +258,7 @@ export default class AppModule { // eslint-disable-line @typescript-eslint/no-ex
 
 export { AppRootComponent as bootstrap }
 export * from './api'
+export { AppHotkeyProvider }
 
 // Deprecations
 export { ToolbarButton as IToolbarButton } from './api'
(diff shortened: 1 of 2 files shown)


Eugeny:
Looking good now, thanks!

PR #8035, opened 2023-03-05
Fix serial Hexadecimal write

### LINKED ISSUE #8034: Serial Hexadecimal Write Issue
**Describe the problem**:
Writing Serialport as Hexadecimal causes wrong data send.

Screenshot 2023-03-05 152659

Here I created a loopback device and send `aa` and receive `ef bd bf`.

I listened hardware with a logic analyzer. The problem is in sending part, not reading. Using other programs then tabby it runs correctly.

**To Reproduce**:
Setup a device with loopback. 
Set input and output as hexadecimal. 
Send `aa`

karaketir16:
fixed #8034

### DIFF
--- a/tabby-serial/src/api.ts
+++ b/tabby-serial/src/api.ts
@@ -132,7 +132,7 @@ export class SerialSession extends BaseSession {
     }
 
     write (data: Buffer): void {
-        this.serial?.write(data.toString())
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you for the fix! @all-contributors please add @karaketir16 for code

Issue #8154, opened 2023-03-28
Search counter starts from 0


princemaple:
**Describe the problem**:
When you search for something, there is a counter that shows you which search result you are on right now.
It should start from 1, but it starts from 0 now. It's quite confusing.

I was searching for something and I was sure I was on the last item, but saw `14/15`. After jumping to the first item and seeing `0/15`, I realized it's counting from 0.

**To Reproduce**:
search anything

PR #8198, opened 2023-04-10
fix Eugeny/tabby#8154 : search panel counter starts from 0

Clem-Fern:
Hi @Eugeny, hope you're doing well :)

This PR aims to fix the issue #8154


Clem-Fern:
Oh yeah absolutely! my bad... I was not seeing the problem like that.
So i think the fix should be `(state.resultIndex ?? 0) + 1`. Currently, it seems that the plus operator only affect the fallback value.

Eugeny:
Right - LGTM now!

Issue #7723, opened 2023-01-03
[feature]add `reconnect tab` to command palette


wangdingwei:
<img width="439" alt="image" src="https://user-images.githubusercontent.com/6716895/210315144-16c0dfbc-046d-4c86-89bf-ad709db5e741.png">
add `reconnect tab` to command palette, thanks

PR #8201, opened 2023-04-11
[feature] add reconnect tab to command palette and hotkey

Clem-Fern:
Hi @Eugeny,

This PR adds reconnect command into the palette like asked in #7723.

In the view to add this feature in the most generic way, I choose to create an interface named `Reconnectable` which Serial, Telnet and SSH tab component are implementing. With this approach, the reconnect command in only added if the current tab can be reconnected. It works as expected but as I am not really familiar with interface in typescript, I let you tell me if it is a correct way to proceed or not...

A more simple way would probably be to provide a context menu for each reconnectable tab (in tabby-serial, tabby-ssh and tabby-telnet).

Secondly, I also added an hotkey to reconnect the current tab. I saw that a restart tab hotkey already exists but it duplicates the current tab and close the old one. Since the restart and reconnect are two different behaviors and reconnect applicable to serial/ssh/telnet only, I thought it could be interesting to have the two options.

I hope i didn't miss anything this time, feel free to ask if any change needed.

### DIFF
--- a/tabby-core/src/configDefaults.linux.yaml
+++ b/tabby-core/src/configDefaults.linux.yaml
@@ -22,6 +22,7 @@ hotkeys:
     - 'Ctrl-Shift'
   duplicate-tab: []
   restart-tab: []
+  reconnect-tab: []
   explode-tab:
     - 'Ctrl-Shift-.'
   combine-tabs:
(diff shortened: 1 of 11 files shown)


Eugeny:
Looking good! At some point in the future I might add a base class for all "connectable" sessions like serial/telnet/SSH but for now this is the best solution

Clem-Fern:
> Looking good! At some point in the future I might add a base class for all "connectable" sessions like serial/telnet/SSH but for now this is the best solution

If you're okay with that, i would like to give a try on this. I think the PR #8206 could be cleaner too with this refactoring.

Eugeny:
For sure!

Issue #8062, opened 2023-03-10
Missing option to automatically close tab when connection is lost


Eugeny:
### Discussed in https://github.com/Eugeny/tabby/discussions/8061

<div type='discussions-op-text'>

<sup>Originally posted by **jarnek73** March 10, 2023</sup>
Hi!

I use tabby to mostly connect to my Ubuntu servers for remote administration.
I miss the option to have tabby automatically close the tab when a connection is lost, eg. when doing a reboot.
So when a ssh-connection is lost, i can choose to let it ask if i want to reconnect like it does now (default), or i can choose to let it close the tab instead (optional).
Hope it makes sense and you will consider this option for a future release.</div>

DerTimonius:
Hi, I would like to investigate this 🙂

andreakeesys:
This can be closed, fixed upstream

Issue #7751, opened 2023-01-06
[feature] Add option to prevent tab from closing when shell/ssh exits


jpbochi:
**Is your feature request related to a problem? Please describe.**

Sometimes, I close a session by accident (e.g. too many ctrl-c's), or kicked off an ssh session. When that happens, the tab closes, and I can't lose all the logs.

Sometimes, I want to finish a session on purpose, but keep the screen log open for reference. That is not currently possible with Tabby, since the tab will be closed.

**Describe the solution you'd like**

I propose that Tabby adopts a similar feature to that iTerm2 has. There, you can choose for each profile, what you want to happen when a session ends, with 3 options:

1. Close tab, which is the default.
2. Kepp tab. A last "session ended" line is printed, no more prompt is asked, but I can still scroll through the logs.
3. Restart session. A "session ended" line is printed, and a fresh shell (or ssh) session starts immediately below it.

Some docs about iTerm2's feature at https://iterm2.com/documentation-preferences-profiles-session.html

Here's an screenshot of their preferences window:

![image](https://user-images.githubusercontent.com/96475/210965001-18d986ca-ab78-40e9-9d21-d7730a51e7e4.png)

**Describe alternatives you've considered**

The only alternative I considered was to configure in my `.bashrc` to prevent bash from closing. This is possible, but it's flimsy and not trivial to do. I don't know if it's possible to do for ssh sessions.

Adding support directly on the terminal app is a far superior option.

rusty-lr:
I'd like to see this also.  There are many advantages to Tabby, but I pop back to mac terminal when I know I'm going to need to review what I was looking at in the ssh session.

Issue #6617, opened 2022-06-15
Auto reconnect on serial port


dave2ip:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Hi there, 
Is it possible to have the serial port automatically reconnect? I am currently connecting an embedded device using a micro USB-D to my PC. This terminal is awesome and will continue to use it even though I have to reconnect manually (press any key).
Thank you
David

PR #8206, opened 2023-04-12
[feature] add configurable behavior when session ends

Clem-Fern:
Hi @Eugeny,

This PR adds different tab's behaviors on session ending.
As asked in #6617, #7751 and #8062, you can choose to keep the tab opened after you got disconnected or exited the shell. You can also choose to automatically reconnect the session or close the tab.

image

I didn't add the ability to configure this feature independently on each profile, but i can definitely take a look if you think it could be useful.

Feel free to ask if any changes are required.

### DIFF
--- a/tabby-serial/src/components/serialTab.component.ts
+++ b/tabby-serial/src/components/serialTab.component.ts
@@ -84,7 +84,7 @@ export class SerialTabComponent extends BaseTerminalTabComponent<SerialProfile> 
         this.attachSessionHandler(this.session!.destroyed$, () => {
             if (this.frontend) {
                 // Session was closed abruptly
-                if (this.config.store.terminal.behaviorOnSessionEnds === 'close') {
+                if (this.config.store.terminal.behaviorOnSessionEnds == 'close') {
                     // Close the tab
                     this.destroy()
                 } else if (this.config.store.terminal.behaviorOnSessionEnds.startsWith('reconnect-or-')) {
--- a/tabby-serial/src/components/serialTab.component.ts
+++ b/tabby-serial/src/components/serialTab.component.ts
@@ -83,12 +83,21 @@ export class SerialTabComponent extends BaseTerminalTabComponent<SerialProfile> 
             this.session?.resize(this.size.columns, this.size.rows)
         })
         this.attachSessionHandler(this.session!.destroyed$, () => {
-            this.write(this.translate.instant(_('Press any key to reconnect')) + '\r\n')
-            this.input$.pipe(first()).subscribe(() => {
-                if (!this.session?.open) {
+            if (this.frontend) {
+                // Session was closed abruptly
+                this.write('\r\n' + colors.black.bgWhite(' SERIAL ') + ` session closed\r\n`)
+
+                if (this.profile.behaviorOnSessionEnd === 'reconnect') {
                     this.reconnect()
+                } else if (this.profile.behaviorOnSessionEnd === 'keep' || this.profile.behaviorOnSessionEnd === 'auto' && !this.isSessionExplicitlyTerminated()) {
+                    this.write(this.translate.instant(_('Press any key to reconnect')) + '\r\n')
+                    this.input$.pipe(first()).subscribe(() => {
+                        if (!this.session?.open) {
+                            this.reconnect()
+                        }
+                    })
                 }
-            })
+            }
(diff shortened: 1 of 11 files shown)


tabby-ssh/src/components/sshTab.component.ts
```diff
@@ -157,24 +157,26 @@ export class SSHTabComponent extends BaseTerminalTabComponent<SSHProfile> implem
     protected attachSessionHandlers (): void {
         const session = this.session!
         this.attachSessionHandler(session.destroyed$, () => {
-            if (
-                // Ctrl-D
-                this.recentInputs.charCodeAt(this.recentInputs.length - 1) === 4 ||
```
Eugeny, line 162: Did this functionality get lost here? For SSH, there's an important distinction: you never want to keep the session open after you explicitly kill it with Ctrl-D or `exit`

tabby-serial/src/components/serialTab.component.ts
```diff
@@ -82,12 +82,25 @@ export class SerialTabComponent extends BaseTerminalTabComponent<SerialProfile>
...
         this.attachSessionHandler(this.session!.destroyed$, () => {
-            this.write(this.translate.instant(_('Press any key to reconnect')) + '\r\n')
-            this.input$.pipe(first()).subscribe(() => {
-                if (!this.session?.open) {
+            if (this.frontend) {
+                // Session was closed abruptly
+                if (this.config.store.terminal.behaviorOnSessionEnds === 'close') {
+                    // Close the tab
+                    this.destroy()
+                } else if (this.config.store.terminal.behaviorOnSessionEnds.startsWith('reconnect-or-')) {
```
Eugeny, line 90: Where is the distinction between `reconnect-or-close` or `reconnect-or-keep-open` is handled? In this case, both fall through into `reconnect()`

tabby-terminal/src/config.ts
```diff
@@ -24,6 +24,7 @@ export class TerminalConfigProvider extends ConfigProvider {
             hideCloseButton: false,
             hideTabOptionsButton: false,
             rightClick: 'menu',
+            behaviorOnSessionEnds: 'keep-open',
```
Eugeny, line 27: Please rename into `behaviorOnSessionEnd`

Eugeny (changes requested):

tabby-ssh/src/components/sshTab.component.ts
Clem-Fern: Oh yeah i can see your point! In that case, wouldn't that be more adequate to create two distinct options :

- Close the tab on session end => only when explicitly terminated (default and actual behaviour) / always / no

- For Serial, Telnet and SSH, automatically reconnect => yes/no *(locked if tabs are always closed on session end)*
(do not try to reconnect if the session has been manually closed even if yes selected)

It would probably be a good comprise for those who want to keep a way to scroll in the log after the session has been terminated (manually or not) and those who want the tabs to close when the connection is lost.

tabby-serial/src/components/serialTab.component.ts
Clem-Fern: Well, for these two choices, i was thinking it as : reconnect ssh/serial/telnet and close or keep open the other kind of session.

tabby-ssh/src/components/sshTab.component.ts
Eugeny: What if we do it the same way iTerm does it with three options (close / keep / reconnect) + a new one ("auto") where `exit` / Ctrl-D is "close" and otherwise "keep"?

And if you implement this on a profile level, then it's possible to set sensible defaults for each session type (close for local, auto for others)

Clem-Fern: Ok let's do this !
I'm converting this PR into draft until it is ready to review again.

Clem-Fern:
Hi @Eugeny :)

As discuss, I've made the changes requested. You can now select 4 different options to choose how the tab should behave when a session ends.

image


I voluntarily set the default mode to `auto` as it is most similar to the actual behavior.

Feel free to ask if any changes are required.

Eugeny (approved):

Eugeny:
LGTM, thank you! I'll keep testing it the next week.

PR #8249, opened 2023-04-17
Update README.ja-JP.md

mottcha:
Hi! I fixed some expressions in Japanese and translated newly texts.
- Fixed some sentences to be more natural and polite expressions
- Added new sponsor keygen


Eugeny:
Thank you!!

PR #8320, opened 2023-04-27
Fix name of scripts in Hacking.md

Kapocsi:
Fixed mismatch in the naming scheme of scripts. 

All the scripts listed under building an installer use the extension `.js` whereas the files themselves use `.mjs`.


Eugeny:
Thank you! @all-contributors please add @Kapocsi for docs

PR #8331, opened 2023-04-28
fix #8297 - set selected text into search box on opening

### LINKED ISSUE #8297: Text selection on terminal not automatically copied when search box is open
**Describe the problem**:
Text selection on terminal not automatically copied when search box is open (alt+f)

**To Reproduce**:
alt+f on a connected host
search some string
select any other string in terminal
Expected: It should be copied automatically to clipboard
Actual: The string in search box is copied to clipboard instead

Clem-Fern:
Hi,

As explained in #8297 , the selected text in terminal is not set into the search box when it opens.
This PR aims to fix this issue.

Fix #8297

### DIFF
--- a/tabby-terminal/src/api/baseTerminalTab.component.ts
+++ b/tabby-terminal/src/api/baseTerminalTab.component.ts
@@ -212,6 +212,11 @@ export class BaseTerminalTabComponent<P extends BaseTerminalProfile> extends Bas
                 this.showSearchPanel = true
                 setImmediate(() => {
                     const input = this.element.nativeElement.querySelector('.search-input')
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks!

Clem-Fern:
> Hi @Clem-Fern - just to clarify the issue [(#8297)](https://github.com/Eugeny/tabby/issues/8297) is not with search-box initialization with the selected text. The problem is when the search box is open (with some text) then if I do a new selection on the terminal window in order to copy it to clipboard, it does not copy the selected text to clipboard rather it takes the text from the search box which is not intended. The search-box text should not interfere with the selection being copied to clipboard.

Oh sorry then... i take a look on it in that way. I will open another PR to address this.

PR #8332, opened 2023-04-28
fix #8297 - prevent copying text when the selection is made by the search box

### LINKED ISSUE #8297: Text selection on terminal not automatically copied when search box is open

Clem-Fern:
Sorry for the inconvenience. This PR comes next to #8331. 

> Hi @Clem-Fern - just to clarify the issue https://github.com/Eugeny/tabby/issues/8297 is not with search-box initialization with the selected text. The problem is when the search box is open (with some text) then if I do a new selection on the terminal window in order to copy it to clipboard, it does not copy the selected text to clipboard rather it takes the text from the search box which is not intended.
The search-box text should not interfere with the selection being copied to clipboard.

Thanks @nexional for the correction. 

This issue seems to only happen when the `Copy On Select` feature is enabled.
The idea here is to prevent the selected text to be copied when the selection is made by the findNext/findPrevious methods.

As I am not really familiar with xterm, I wasn't able to find if there's a way to prevent this from happening directly through the search addon api. Feel free to ask me if any changes are required.

Fix #8297

### DIFF
--- a/tabby-terminal/src/frontends/xtermFrontend.ts
+++ b/tabby-terminal/src/frontends/xtermFrontend.ts
@@ -69,6 +69,7 @@ export class XTermFrontend extends Frontend {
     private resizeHandler: () => void
     private configuredTheme: ITheme = {}
     private copyOnSelect = false
+    private preventNextOnSelectionChangeEvent = false
     private search = new SearchAddon()
     private searchState: SearchState = { resultCount: 0 }
     private fitAddon = new FitAddon()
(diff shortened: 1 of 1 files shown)


Eugeny:
I think that's still not quite what @nexional meant, unless I'm misinterpreting it, it's about being unable to "select-to-copy" while the search box is visible, correct me if I'm wrong

Clem-Fern:
Well, very sory again then...

I've just tested it another time. I was able to select-to-copy with the search box opened. However, I noticed that the findNext and findPrevious methods raised an onSelectionChange event.

As there's no distinction if the selection is made by the user or the search box in `this.xterm.onSelectionChange`, it replaces the previously copied tex in the clipboard by the value entered in the search field.

Eugeny:
No worries! I think this is still a useful change anyway!

PR #8471, opened 2023-05-22
Fire manual keyup event when key is not Enter

### LINKED ISSUE #8071: Switching pane-maximize will add an additional line break in the new window
**Describe the problem**:
Switching pane-maximize will add an additional line break in the new window, which can be solved by changing a shortcut key, but I'm still used to the combination of cmd option enter.

image

**To Reproduce**:
split window>cmd option enter * n

echo304:
This is experimental change to fix #8071 

By adding a condition to check keyname for manual keyup event firing,
the bug is fixed but the side-effect is unpredictable as of now.

### DIFF
--- a/tabby-core/src/services/hotkeys.service.ts
+++ b/tabby-core/src/services/hotkeys.service.ts
@@ -178,7 +178,7 @@ export class HotkeysService {
             this._key.next(getKeyName(eventData))
         })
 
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks, let's see how it performs. @all-contributors please add @echo304 for code

Issue #7390, opened 2022-10-30
Fast launch


bigwhite2022:
In some low configuration computer,this app launched toooooo slow,My computer is intel i5&1050Ti, it launch about 20s more!!
I think it can add a option, "click close button to taskbar lower right corner small icon",I think it perhaps launch faster.

English not well..Forgive me

Hope to be adopted,Thanks

F3000:
or add an option: start with windows (but minimized in taskbar)

Issue #6295, opened 2022-04-27
it is too slow to start the tabby


czy2014hust:
Version: 1.0.176
Platform: macOS x64 21.4.0
Plugins: plugin-fig-integration
Frontend: xterm

jenningszheng:
yes, I feel the same.

JoshTseeee:
> yes, I feel the same. Compared with iTerm, tabby is slower, which I can tell with my eyes.

+1

dnut:
Yeah same deal on linux. It takes several seconds to start, which is annoying for quick tasks where I'm just looking to open a new window, run a couple commands, and close it. Currently tabby is not really appropriate for this workflow because it's so slow to open and cumbersome to close.

Margrie:
I feel the same, can someone tell why tabby starts so slow compared to the normal termininal like xterm or GNOME-Terminal. Is there an option to start tabby faster?

dnut:
> Is there an option to start tabby faster?

Yeah! That's easy. Just use hyper, tilix, wezterm, or terminator instead of tabby. It takes 0 seconds to start if you never start it.

Issue #6253, opened 2022-04-23
Startup too slow [Suggestion]


williamgh2019:
Version: 1.0.176
Platform: macOS x64 21.4.0
Plugins: none
Frontend: xterm

It could be better if we can startup Tabby faster like Mac Terminal.

MANICX100:
Agreed, but you must understand Tabby is a larger application than the built in terminal prompt both in features and memory usage.

Personally I use quake mode (quick summon) to have it instantly appear and never close it so the slow initial startup doesn't affect me.

You will be prompted to use ctrl and space on first launch, you can get this hotkey under settings, config.

TisnKu:
Opening new tabs and closing tabs are slow. When tab is being opened or closed, there is this animation of title collapse/extend which looks laggy. Maybe we can remove the animation?

amilor:
I have noticed this lagginess when opening/closing tabs after I updated to 1.0.176. Disabled it by turning off the Settings > Application > Accessibility > Enable animations toggle.

williamgh2019:
> Agreed, but you must understand Tabby is a larger application than the built in terminal prompt both in features and memory usage.
> 
> Personally I use quake mode (quick summon) to have it instantly appear and never close it so the slow initial startup doesn't affect me.
> 
> You will be prompted to use ctrl and space on first launch, you can get this hotkey under settings, config.

If you compare startup with Termius SSH you will see, Tabby startup is really slow and Termius start very fast. If Tabby can fix this issue it could be better.

PR #8528, opened 2023-06-02
Improving tabby startup time

Clem-Fern:
Hi @Eugeny,

I found some issues about how tabby can be slow to launch in some case (#6295 , #7390 , #6253 ).
After taking a quick look, I saw that installed plugins are loaded one after another at the app startup. 

This PR aim to improve the startup time by finding and loading the plugins simultaneously.

Startup on latest release with 0 plugins : ~2.9sec
Startup on latest release with 8 plugins : ~3.5sec

Startup on this PR with 0 plugins : ~2.3sec
Startup on this PR with 8 plugins : ~2.3sec

I've only have SSD so there is no huge gab on my side. (Tested on Debian 11, don't have other system on hand sorry :/)

As always, I am not sure theses changes are really relevant nor if it is a good approach to try to reduce the startup time. I hope you could give me a feedback on this when you got time :D

### DIFF
--- a/app/src/plugins.ts
+++ b/app/src/plugins.ts
@@ -255,8 +255,8 @@ export async function loadPlugins (foundPlugins: PluginInfo[], progress: Progres
             setTimeout(x, 50)
         }))
     }
-    progress(1, 1)
-
     await Promise.all(pluginsPromises)
+
+    progress(1, 1)
     return plugins
 }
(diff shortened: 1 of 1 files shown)


Eugeny:
I've tested it and it improves my startup time from 2.7s to 2.1s, great work!

Clem-Fern:
Good to hear that !

Clem-Fern:
Well, I took another look in it to be sure I had done everything I could. But I think we're good for this time.

I've just added a new commit to fix a mistake I've made in the placement of the call of the progress bar.

Eugeny:
Perfect 👌

PR #8612, opened 2023-06-23
Update settingsTabBody.component.ts

qyecst:
Increase the value of max-width to avoid shadowing the font size in the Settings (Settings - Appearance - Font).

The font size is 14, but only 1 can be seen.
图片
图片

After increase the max-width to 600px.
图片

The reason for choosing 600px is that I saw that the max-width of div below is 600px.
图片


Eugeny:
Thank you for the contribution! @all-contributors please add @qyecst for code

qyecst:
Hello @Eugeny Looks like the branch all-contributors/add-qyecst doesn't update all README files properly, maybe it should be deleted.

Eugeny:
Should be fixed now :)

Issue #8534, opened 2023-06-05
The settings button doesn't respond sometimes


shankru5:
Version: 1.0.197
Platform: Windows x64 10.0.22000
Plugins: save-output
Frontend: xterm-webgl

I have noticed that the settings button stops responding or doesn't open the settings tab as expected sometimes when I click on the gear icon on the top right. I am not exactly sure I have found a way to reproduce the issue consistently. But it does happen from time to time.

steberk:
I found a way to reproduce (maybe there is other one too)
- Start Tabby
- Open a SSH connection
- Split to the bottom (you can drag and drop other connection too)
- Save the layout from the tab context menu 
- Now the settings tab cannot be open (gear icon, setting menu from the main page, hotkey...)

As I remember when I played with multiple panes in multiple tabs this problem appeared without saving layout. But I could not reproduce this time...

shankru5:
> I found a way to reproduce (maybe there is other one too)
> 
> * Start Tabby
> * Open a SSH connection
> * Split to the bottom (you can drag and drop other connection too)
> * Save the layout from the tab context menu
> * Now the settings tab cannot be open (gear icon, setting menu from the main page, hotkey...)
> 
> As I remember when I played with multiple panes in multiple tabs this problem appeared without saving layout. But I could not reproduce this time...

You are right. I followed your steps and it is reproducible now for me as well.

steberk:
Hello,

I've attached the log. The first exceptions were raised when I saved the
layout. The last exception was raised when I pressed the settings button.

Eugene ***@***.***> ezt írta (időpont: 2023. júl. 4., K,
15:28):

PR #8663, opened 2023-07-07
Feat#5688: Save As Profile option available for all tabs

### LINKED ISSUE #5688: Save Connection As missing on SSH tabs
### Discussed in https://github.com/Eugeny/tabby/discussions/5686

<sup>Originally posted by **ismaelbalkhi** February  7, 2022</sup>
Hello,

I love Tabby! Now that that is out of the way. I constantly SSH into hundreds of hosts, it would be nice if I could **'Right-Click'** on an established connection and **'Save-As'** a profile along with the login credentials I used. Right now, this easy method is not available. I have to go, one-by-one into Settings to create individual profiles which isn't feasible in the long run for so many connections. Also, does anyone know of a nice Sessions Manager within Tabby that could handle hundreds of saved profiles and be able to organize/sort them? 

Currently, I quickly connect via myName@10.10.10.10:22, but I don't see a way to save it efficiently.

Thanks!</div>

Clem-Fern:
Hi @Eugeny, I hope you are doing well !

I took a look on the feature request #5688 . This PR aims to make the 'Save As Profile' context menu item available for all tabs.

1. I saw that they were no ID setted with the new saved profile before. It causes the profile to not be shown in the `Profiles & Connections` selector since the commit c983743 : 
https://github.com/Eugeny/tabby/blob/c983743b57d90246d997b1ebf328a5c6696ac315/tabby-core/src/services/profiles.service.ts#L156
Should we create a recovery mechanism? Like creating an ID for profile those does not have one when the config is saved. What would be your approach on this?

2. During my test, I encountered the same issue as the one describe in #8534. https://github.com/Eugeny/tabby/blob/c983743b57d90246d997b1ebf328a5c6696ac315/tabby-core/src/services/config.service.ts#L200-L210 Even with `JSON.parse(JSON.stringify(this._store))` and in rare case, some invalid yaml type seems to be present in the config object on yaml dump. Sadly I wasn't able to identify which part precisely... Anyway, adding `skipInvalid: true` in yaml dump option prevent the method to rise `{name: 'YAMLException', reason: 'unacceptable kind of an object to dump [object Function]'}` by skipping invalid type.

3. Also, It could be great to do a bit of refactoring in the future and create methods in the ProfileService for profile creation, edition and deletion to have only one entrypoint for profile operation. It would be simpler to maintain. What do you think ?

As always, feel free to ask me if changes needed :)

Resolve #5688

### DIFF
--- a/tabby-core/src/services/config.service.ts
+++ b/tabby-core/src/services/config.service.ts
@@ -205,7 +205,7 @@ export class ConfigService {
         // Scrub undefined values
         let cleanStore = JSON.parse(JSON.stringify(this._store))
         cleanStore = await this.maybeEncryptConfig(cleanStore)
-        await this.platform.saveConfig(yaml.dump(cleanStore))
+        await this.platform.saveConfig(yaml.dump(cleanStore, {skipInvalid: true}))
         this.emitChange()
     }
 
@@ -213,7 +213,7 @@ export class ConfigService {
      * Reads config YAML as string
      */
     readRaw (): string {
-        return yaml.dump(this._store)
+        return yaml.dump(this._store, {skipInvalid: true})
     }
 
     /**
--- a/tabby-core/src/services/config.service.ts
+++ b/tabby-core/src/services/config.service.ts
@@ -205,7 +205,7 @@ export class ConfigService {
         // Scrub undefined values
         let cleanStore = JSON.parse(JSON.stringify(this._store))
         cleanStore = await this.maybeEncryptConfig(cleanStore)
-        await this.platform.saveConfig(yaml.dump(cleanStore, {skipInvalid: true}))
+        await this.platform.saveConfig(yaml.dump(cleanStore, { skipInvalid: true }))
         this.emitChange()
     }
 
@@ -213,7 +213,7 @@ export class ConfigService {
      * Reads config YAML as string
      */
     readRaw (): string {
-        return yaml.dump(this._store, {skipInvalid: true})
+        return yaml.dump(this._store, { skipInvalid: true })
     }
 
     /**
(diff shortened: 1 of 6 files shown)


Eugeny:
Wow - great job! Thank you for your continued efforts improving the app by the way :)

1. We can just assign random UUID IDs to any profiles that don't have one by adding a migration to [ConfigService](https://github.com/Eugeny/tabby/blob/master/tabby-core/src/services/config.service.ts#L282)
2. Weird - the only way would be if some non-json-safe values are added again by `maybeEncryptConfig` ? Can't immediately think of any place this would happen in
3. Sounds great!

Clem-Fern:
Thank's for your quick feedback. Well, I'm really glad to know that I'm able to help a bit with my small skills. I would like to help more but the greatest majority of issues stills absolutely out of my league for now ahah

1. Noted, I will add a commit for that in the next few days.

2. I think it stills a good idea to leave the `skipInvalid: true` option to have an additional safeguard. Be sure that I will take another look into this too.

3. Good! Expect a future PR ;)

Clem-Fern:
Sorry for the delay,

1. As discuss, I added a new config migration to assign IDs to any profiles that don't have one. (Lint Job failed during Install Deps)
2. It's really weird... I was able to reproduce the issue #8534 almost all the time by following the path describe by @shankru5. Even with the code below in the `save` method, the YAMLException still rise. However, it seems that we do not encounter any error if the writing/parsing of the yaml is done with the [yaml](https://github.com/eemeli/yaml) lib.
```javascript
 async save (): Promise<void> { 
     await lastValueFrom(this.ready$) 
     if (!this._store) { 
         throw new Error('Cannot save an empty store') 
     } 
     // Scrub undefined values 
     let cleanStore = JSON.parse(JSON.stringify(this._store)) 
     // cleanStore = await this.maybeEncryptConfig(cleanStore) 
     await this.platform.saveConfig(yaml.dump(cleanStore)) 
     this.emitChange() 
 } 
```

Ready to merge from my side if you are ok with .1 and with the `skipInvalid: true` workaround for the .2.

Issue #5773, opened 2022-02-16
Telnet Echo,when i input a char,it give me double char


huawang400:
dear:
system:window64
version:1.0.172 

Telnet when i input a word ,he give me doble char,
but when i connet ssh,it's ok

![image](https://user-images.githubusercontent.com/75016485/154191629-0cedc274-b19a-45b0-8163-25be51a06d1c.png)

![image](https://user-images.githubusercontent.com/75016485/154191850-083d532b-6b0e-4649-8840-d62e656465c9.png)

thank you for focusing on this problem!!!

Eugeny:
Could you please record the telnet session via Wireshark and post it here or send it to me at e@ajenti.org?

DehanLUO:
I met the same problem. Version 1.0.197 windows.
I added some log at
```js
// ./Tabby/resources/builtin-plugins/tabby-telnet/dist/index.js
   processTelnetProtocol(data) {
      this.logger.info(data); // LOG ADDED
        while (data.length) {
            if (data[0] === TelnetCommands.IAC) {
                this.logger.info(TelnetCommands[data[0]] +" " + TelnetCommands[data[1]] + " " + TelnetOptions[data[2]]); // LOG ADDED
   ...
   }
```
And here is what I got
<img width="1406" alt="image" src="https://github.com/Eugeny/tabby/assets/53093688/be2d6f30-e0d3-4884-9f88-66234b3bd448">

It seems that `IAC DO ECHO` requested that the receiver of this command begin echoing, making `this.streamProcessor.forceEcho = true`. Then, `IAC WILL ECHO` requested to begin echoing data characters the server received over the telnet connection back to the sender of the data characters.

Issue #5042, opened 2021-12-06
Double character display in telnet mode


huangjiaming016:
Version: 1.0.165-nightly.0
Platform: Windows x64 10.0.19041
Plugins: none
Frontend: xterm

Double character display in telnet mode

Eugeny:
Could you please attach a Wireshark dump of the connection? (or send it to e@ajenti.org)

sasa9451:
I'm experiencing the same issue. Does anyone have any suggestion?
version:1.0.169
Platform windows x64 10
Frontend: xterm

huawang400:
Platform windows x64
I also have this problem,
first i think it is my version,but i already have updated ,it didn't solve this problem

pictureshot like this
![image](https://user-images.githubusercontent.com/75016485/154202981-0bccccb7-54db-46c2-b18d-83b3321c092d.png)
![image](https://user-images.githubusercontent.com/75016485/154203016-cf540ee7-45ad-4614-9032-fe8e6c9dafa0.png)

PR #8677, opened 2023-07-10
Stop local echo when remote echo is on

DehanLUO:
To solve the [issue#5773](https://github.com/Eugeny/tabby/issues/5773) and the [issue#5042](https://github.com/Eugeny/tabby/issues/5042).


Eugeny:
Thank you! @all-contributors please add @DehanLUO for code

Issue #7199, opened 2022-09-30
the windows controlls moves to left bottom when tabs are aranged to left.


rahulsinghk:
Version: 1.0.183
Platform: Windows x64 10.0.19044
Plugins: title-control
Frontend: xterm

when we arrange the tabs to the left of the pane, the windows control where we minimize, maximize, and cut the application moves to the bottom left. 

![image](https://user-images.githubusercontent.com/42302337/193184231-c2d38408-c326-4c17-93ed-101b71eda4ac.png)

Eugeny:
Thanks. As a workaround, you can set Settings -> Window -> Titlebar to "full" or "native"

rahulsinghk:
Thanks for the workaround,
one more small issue, 

the tab's width dynamic text is not showing properly.  

![image](https://user-images.githubusercontent.com/42302337/194257903-3f395a28-589e-46a1-ac3e-a6706171f54e.png)

Issue #8650, opened 2023-07-05
Main window title bar changes its position when switching tabs location


cichy19:
Version: 1.0.197
Platform: Windows x64 10.0.19045
Plugins: none
Frontend: xterm-webgl

**Description:**
When changing Tabs location (Settings/Window/Tabs/Tabs location) to Bottom, Left or Right then main window title bar changes its position and appears on the bottom.

**Expected behaviour:**
Main window title bar does not change its position when changing tabs location.

![tabby_tabs_location_right](https://github.com/Eugeny/tabby/assets/31633602/eda618df-4711-4615-b6d9-0aec08273d37)
![tabby_tabs_location_left](https://github.com/Eugeny/tabby/assets/31633602/466ed6bb-424f-4811-99cb-56aeeecd088e)
![tabby_tabs_location_bottom](https://github.com/Eugeny/tabby/assets/31633602/712ffd4e-1ff4-47e1-8e98-c7f1ef9a218e)
![tabby_tabs_location_top](https://github.com/Eugeny/tabby/assets/31633602/2ebb6d8c-5017-4114-95ee-fed78d545b92)

Issue #6278, opened 2022-04-26
UI bug: Min/Max/Close buttons in weird place (Fedora 35)


SolidState214:
See screenshot below,, buttons are obscured and placed at bottom left of the window. On Fedora 35 with window frame set to thin.

![tabby_ui_bug](https://user-images.githubusercontent.com/11671541/165216408-0619a933-d24a-4eec-b16f-409f619f0d84.png)

Issue #4292, opened 2021-07-29
Window frame header missing when using left/right tab location


mullerch:
When configuring the tabs to be at left/right location, it the whole window frame header (the top bar of the window where there is generally the application name and the window buttons) that moves to the side.

This leads to having the minimize/maximize/close buttons at the bottom of the tabs selector instead of the top-right corner of the window (as usual). Also, to move the window, one must grab it by the tab selector empty space, which is not intuitive.

I suggest to split the location of the tab selector and the window frame header. The window frame header should always sit on top and the tab selector should be configurable as it is now.

Eugeny:
You could keep the tabs on the side but still have a normal window frame on the top by setting `Window frame` to `Native` in window settings

mullerch:
That's a nice workaround, strongly lowering the priority of a nicer fix.

PR #8707, opened 2023-07-14
Implement Window Controls Overlay

FeroTheFox:
Fixes: #7199 #8650 #6278 #4292 (on Windows)

Hey @Eugeny, I've daily-driven Tabby for a little while and I love its customizability. It reminds me of Vivaldi in a lot of ways! This is my first PR here and I'd love to try to fix a long-standing issue with the tab bar location and the window controls.

This PR aims to mitigate this issue by replacing the custom window controls with the Window Controls Overlay API that Electron provides.

Specifically, my code:
- Updates Electron to the latest stable (from version 22 to version 25) to fix an issue where overlay buttons with transparennt backgrounds have issues with hover styles (see https://github.com/electron/electron/pull/38693). This does not seem to affect any other functionality in Tabby - all packages appear to work fine and no obvious breaking changes were observed. I've been able to build and run the app perfectly fine.
- Adds new styles to compensate for the space taken up by the controls. Extra padding and adjustments are needed to prevent clipping.

## Known issues
I've tested this throroughly on Windows 11, but there's some other issues with this implementation that need to be addressed in order for this to be considered production-ready. 

### Issue 1
- Currently, this is only feasibly usable dark themes. The background color of the overlayed buttons are transparent to not clash with any themes, but the icon colors are hardcoded to white.

here's a visual example:
image

*Dark themes work flawlessly, however in lighter themes...*

image
It is impossible to know the window buttons exist, even when hovering directly over them.

We'd need to implement a way to pass the currently applied theme's colors to Electron (probably via IPC) and update the colors of the window controls accordingly so that they don't cause issues like this.

### Issue 2
- the settings button is obscured when the window frame is set to thin and tabs location is on top.
image

### Issue 3
- I did not consider Linux heavily when I made this PR. Linux users will still see a broken window controls experience when not using the native titlebar. This should probably be set by default if a user is on Linux.

Other than that, I'm proud of how this turned out!
electron_un4bK8oc2L


Eugeny:
Thank you! Might be a stupid question, but what are the advantages of WCO besides it being native? 

PS issue 1: you'd need to watch for config changes and respond to them in `window.ts` same way it's done for window vibrancy settings right now. Issue 2 is just a missing margin somewhere now that the window controls are not in the DOM tree.

FeroTheFox:
Not a stupid question at all! That's exactly its advantage. It unifies window control implementations across macOS and Windows. IIRC setting `frame` to false on macOS defaults to WCO. On Windows 11, it lets people use snap layouts

FeroTheFox:
@Eugeny Pinging for the latest update - the WCO colors now respect the theme that you're using :)

https://github.com/Eugeny/tabby/assets/52982404/7ea77977-5a90-41fc-bace-e8ce9e15aab8

FeroTheFox:
After a day of using a built version of the last commit here, I can say it's working quite smoothly now. Not sure how to resolve the CI errors, but I'm sure it's nothing a re-run can't fix. No rush to merge or anything :)

app/lib/window.ts
```diff
@@ -94,12 +94,19 @@ export class Window {
         } else {
             if (process.platform === 'darwin') {
                 bwOptions.titleBarStyle = 'hidden'
+                // If not macOS and native appearance is not toggled, use WCO.
+            } else {
```
Eugeny, line 98: Does WCO work on Linux? This code path implies it does, the comment in [titleBar.component.scss](https://github.com/Eugeny/tabby/pull/8707/files#diff-d148ef5900e1ff0d7deaf652d073de91b29f7bf9c814006c71fda13d72acc853) implies it doesn't.

app/lib/window.ts
```diff
@@ -94,12 +94,19 @@ export class Window {
...
+                bwOptions.titleBarStyle = 'hidden'
```
Eugeny, line 99: can be moved out of the `if`

app/lib/window.ts
```diff
@@ -94,12 +94,19 @@ export class Window {
...
+                bwOptions.titleBarOverlay = {
+                    color: '#00000000',
+                }
             }
         }
 
         if (process.platform === 'darwin') {
             this.window = new BrowserWindow(bwOptions) as GlasstronWindow
         } else {
+            // this.window = new BrowserWindow(bwOptions) as GlasstronWindow
```
Eugeny, line 109: Please clean up commented code across the PR

tabby-core/src/components/appRoot.component.pug
```diff
@@ -1,16 +1,20 @@
 title-bar(
   *ngIf='ready && !hostWindow.isFullscreen && config.store.appearance.frame == "full" && config.store.appearance.dock == "off"',
   (dblclick)='hostWindow.toggleMaximize()',
+  [class.hide-controls]='hostApp.platform !== Platform.Linux && !hostWindow.isFullscreen',
```
Eugeny, line 4: should be controlled via an @Input on TitleBarComponent, not a class

Eugeny (changes requested):
Added a few comments, otherwise LGTM :v:

app/lib/window.ts
FeroTheFox: [It doesn't](https://www.electronjs.org/docs/latest/tutorial/window-customization#window-controls-overlay-macos-windows), but good catch, will fix

tabby-core/src/components/appRoot.component.pug
FeroTheFox: Done in the latest commit, window controls in the titlebar component is an ngIf now ✨

FeroTheFox:
I believe that resolves your comments :)

Eugeny:
Tested on all 3 platforms and it works nicely - well done! 

@all-contributors add @FeroTheFox for code

Clem-Fern:
Hi @Eugeny, @FeroTheFox :)
I hope you two doing well.

I encountered an error while working on tabby with the master branch and I think it has something to do with this PR (I'm on Linux):
```
index.ts:173 Unhandled exception: TypeError: this.window.setTitleBarOverlay is not a function
    at IpcMainImpl.<anonymous> (/home/clem/git/tabby/app/dist/webpack:/tabby/lib/window.ts:397:25)
    at IpcMainImpl.emit (node:events:513:28)
    at IpcMainImpl.emit (node:domain:489:12)
    at WebContents.<anonymous> (node:electron/js2c/browser_init:2:88498)
    at WebContents.emit (node:events:513:28)
    at WebContents.emit (node:domain:489:12)
```

https://github.com/Eugeny/tabby/blob/0101ffddb65b6fac3553499d7e2d113f31d23ff1/app/lib/window.ts#L390-L403

After a quick look in the electron [doc](https://www.electronjs.org/fr/docs/latest/api/browser-window#winsettitlebaroverlayoptions-windows), it seems that the function `setTitleBarOverlay` is available on Windows platform only.

Eugeny:
Hey, normally when Electron docs refer to functions as only available on platform X, that means it's a noop on other platforms (also supported by the electron TS typings). I've seen it too, but the error by itself is harmless as the setTitleBarOverlay call is at the very end of the event handler.

I'm waiting to see if it's going to be fixed in the electron updates and if it's not, I'll just replace it with `this.window['setTitleBarOverlay']?.(...`

Clem-Fern:
Oh I see, I was not sure what was the impact exactly so, in doubt, i preferred report this.
Thank's for your insight anyway ;)

PR #8709, opened 2023-07-15
Feat#8680: add "connect to" option in profile selector for each provider supporting quick connect

### LINKED ISSUE #8680: Add "connect to" option to type-specific profile selectors
### Discussed in https://github.com/Eugeny/tabby/discussions/8679

<sup>Originally posted by **BallisticCrux** July 10, 2023</sup>
Thought I would be able to type "telnet 1.2.3.4" into the quick selector or any other keyword but every time it defaults to connecting via SSH. I've looked through the documentation and cannot find if this is a feature or if its not supported and I have to manually create a specific profile for every connection type other than SSH. Attempted asking for clarification on an open query but it was not elaborated on.

image
</div>

Clem-Fern:
Hi @Eugeny :)

As discuss in #8680, this PR aims to add "connect to" option in profile selector for each provider supporting quick connect:
image

Should we create a configuration to put weight on "connect to" option? It would allow users to choose which "Connect to" option have to be on top of the list according to their habit.

Feel free to ask me if changes needed!

Resolve #8680

### DIFF
--- a/tabby-core/src/components/selectorModal.component.ts
+++ b/tabby-core/src/components/selectorModal.component.ts
@@ -76,10 +76,11 @@ export class SelectorModalComponent<T> {
                 { sort: true },
             ).search(f)
 
-            const freeOption = this.options.find(x => x.freeInputPattern)
-            if (freeOption && !this.filteredOptions.includes(freeOption)) {
-                this.filteredOptions.push(freeOption)
-            }
+            this.options.filter(x => x.freeInputPattern).forEach(freeOption => {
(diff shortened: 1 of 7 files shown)


Eugeny:
I think that's a good idea seeing as how currently Telnet ends up on top :D 
We don't need a full priority reorder UI, just a single option to select the preferred "default" connect-to provider is enough (you can put it in Settings -> Profiles -> Advanced, it should be SSH by default)

Clem-Fern:
As discuss :)
image

Not quite sure about the name and description I should add for this option by the way '^^

Eugeny:
Nice!

PR #8719, opened 2023-07-18
Adding an introduction to sftp tab, people who switch from SecureCRT to Tabby may want to use similar features of sftp tab.

wljince007:
Adding an introduction to sftp tab, people who switch from SecureCRT to Tabby may want to use similar features of sftp tab.

### DIFF
--- a/README.de-DE.md
+++ b/README.de-DE.md
@@ -119,6 +119,7 @@ Plugins und Themen können direkt aus der Ansicht "Einstellungen" in Tabby insta
 * [clippy](https://github.com/Eugeny/tabby-clippy) - ein Beispiel-Plugin, das einen die ganze Zeit nervt
 * [workspace-manager](https://github.com/composer404/tabby-workspace-manager) - ermöglicht das Erstellen eigener Workspace-Profile auf Basis der angegebenen Konfiguration
 * [search-in-browser](https://github.com/composer404/tabby-search-in-browser) - öffnet den Standard-Systembrowser mit einem Text, der aus dem Tabby Tab ausgewählt wurde
+* [sftp-tab](https://github.com/wljince007/tabby-sftp-tab) - open sftp tab for ssh connection like SecureCRT
 
 <a  name="themes"></a>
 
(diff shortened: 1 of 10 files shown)


Eugeny:
Heya I've just looked at your plugin code - good idea! However I would avoid placing profile templates into defaults since this will lead to confusing behaviour when a user already has existing profiles. A much cleaned solution would be to implement a ProfileProvider and provide your three templates (or automatically select one correct template depending on the platform!) as built-in profiles via `getBuiltinProfiles` like here: https://github.com/eugeny/tabby/blob/master/tabby-ssh/src/profiles.ts#L78C17-L98

wljince007:
Thank you for your patient guidance. I will try to modify the plugin code according to your guidance。

wljince007:
@Eugeny The profile for the three platforms I provide here is a template. If it is provided as a builtin profile in the form of a ProfileProvider, users also need to copy and create a new profile to modify the sftp path. Once copied, the sftp tab plugin cannot know the profile name and cannot use this profile when creating sftp tab.

I see an isTemplate attribute in the profile. If this is set be true, the templates of the three platforms will not appear in the user's selection of "Profiles&Connections". However, the settings can be edited according to the local profile. This is currently the best implementation within my capabilities, and it is the simplest way for the sftp tab plugin to maintain, keep the minimum coupling between sftp-tab and tabby.

There are still many people who want the functionality of the full keyboard operated sftp. From the perspective of plugin download volume page:
```
https://npm-stat.com/charts.html?package=tabby-sftp-tab&package=tabby-trzsz&package=tabby-docker&package=tabby-serial-timestamp&package=tabby-workspace-manager&from=2023-07-11&to=2023-08-11
```

As a concept, functions of sftp-tab plugin is expected to be implemented by you(Tabby) by default, so that it can be used across the entire platform.

wljince007:
@Eugeny   Actually, ssh2sftp_XXX_template is intended to be used as a template, the main goal is not allow user to create SFTP profiles,  the initial idea was let sftp-tab plugin uses this template to create  sftp connections of ssh  automatic. The ssh2sftp_XXX_template are provided in the form of local profile template, just want to borrow the interface modification function of the localProfile in the tabby-local to let users modify the sftp path and save it in the config file. 

(其实三平台的 ssh2sftp_XXX_template就是想做为模板来使用，最开始的想法是让sftp-tab插件使用这个模板来创建ssh的sftp连接。以localProfile template的形式提供三模板，是想借用tabby-local中localProfile的界面修改功能，让用户修改其中的sftp地址，并保存在配置文本中)

wljince007:
@Eugeny  From the perspective of plugin downloads, people still prefer to use the functionality of sftp tab

https://npm-stat.com/charts.html?package=tabby-sftp-tab&package=tabby-trzsz&package=tabby-docker&package=tabby-serial-timestamp&package=tabby-workspace-manager&from=2023-07-11&to=2023-08-11

image

Eugeny:
LGTM but please make sure you're not making other unrelated changes to the READMEs

PR #8416, opened 2023-05-12
Refactoring connectable tab

Clem-Fern:
Hi @Eugeny, hope you're doing well

As discuss in #8201, I've tried to refactor ssh, serial and telnet tab component and created a connectable base class.
All commits were made and tested meticulously to try to avoid any breaking change.

I voluntarily left the `Reconnectable` interface but it can probably be deleted. Same for `restart-telnet-session`, `restart-telnet-session` and `restart-telnet-session`, since there are a global `reconnect` hotkey for connectable tab.

I won't sure there be any good to take in this PR but I have done my best ahah :D
Could you, please, give me a feedback on this?

Issue #6650, opened 2022-06-22
Feature request: Add possibility that all profiles from group can be opened at once


Bootta11:
Use case: for example you are working on different projects where every project had need to have multiple tabs opened(running commands, logs, monitor system etc). To make easier to switch from project to project make feature to define multiple tab groups per project and open it by 1 click on selected group.

Solution 1: Is it possible to add possibility to open all profiles from group by clicking on group name. Like we had already option to select profile and open new tab with it, just add possibility to click on group name and open tabs for all profiles in group. 

Solution 2: You can maybe add another feature to define some kind of open tab groups so instead of opening all profiles from profile group we will open all profiles from open tab group(this is better because you can define to open some profile or group 2 times or you can define some other custom options). I guess first suggestion is lot easier to implement but this is more customizable.

JakeyPrime:
I run a Cluster of 10 nodes in a testing lab for work and this would be a godsend, we have multiple clusters of 10 nodes that all need to be addressed and "broadcasted" to. This would be excellent to have.

davixd:
I would like to have this feature as well.

171121130:
me too!

whysi:
Same here, there is also the issue https://github.com/Eugeny/tabby/issues/8157 that refers something similar.
A feature like "open all profiles in this group" would be immensely helpful from a sysadmin point of view.

Jackysi:
I agree. I found Tabby recently and I cannot express how much love I have for this app. It's awesome. There are some features I'd love from Xshell, like Open Group, Send commands to group, etc... Bit offtopic apologies, but I'd like to see this feature as well, it would really make sense for sysadmins

zz0rk:
Is this feature implantation still under consideration? Thank you

Issue #3880, opened 2021-05-21
[feature request] Sessions sorting and tree hierarchy


eugenov:
Version: 1.0.139
Platform: win32 10.0.19042
Plugins: save-output

Hello!
Is there any plans to implement ssh sessions sorting in any way (besides manual config reordering)? Drag'n'drop + tree structure will be very helpful.
Thank you!

robpomeroy:
+1 🙂 - I pedantically sort almost everything, alphabetically. I have about 30 profiles configured. Would love to be able to sort them conveniently.

PR #8726, opened 2023-07-20
Various fix/enhancement about Profiles

### LINKED ISSUE #3999: Allow SSH groups to specify settings that hosts inherit

### LINKED ISSUE #7057: Last profile auto select
Can we have the last (recently connected) connection/profile selected in the profile selection dialog? Currently it by default selects the topmost one.

### LINKED ISSUE #7265: Resetting specified profile default?
Request for a method to reset specified default profile settings.
Sometimes I can't make sure if any wrong setting options are made in the profile default(e.g. SSH default), so it'd be better to set up a 'reset' button.

Thx!

### LINKED ISSUE #8065: Click '+' won't open new session in one situation
Version: 1.0.189
Platform: macOS x64 22.3.0 (Ventura 13.2.1)
Plugins: none
Frontend: xterm-webgl

I encountered a problem with opening a new session by clicking '+' button. This happens on both Mac OS and Windows 10/11. Here is how to reproduce it.

1. Click session selector menu (20+ pre-defined sessions in my case), don't choose any selection. Just click any empty space in tabby. The selector menu goes away as expected.

2. Now click '+' button or '+ New terminal'. This will no longer trigger an opening of a new session (which is supposed to). But if you click any empty space in tabby, the session will then be opened up. It feels like the click to '+' is stacked up until you click an empty space in tabby. With that said, if you click '+' 3 times (nothing happens) and then click an empty space in tabby, 3 sessions will be opened up all at once.

3. This behavior can be reset back to normal by restarting tabby but it's kinda annoying.

Thanks.

Clem-Fern:
Hi @Eugeny, hope you're doing great !

I'm currently working on a small list of issues which are related to Profiles and that I want to try to address. I'm opening this draft to keep track on my progress, and I will probably ask for your opinion on the way to implement some feature if you are ok with that.

Resolve #7057 4dedbbc25a8f09d69630bbe9de81be5824fa0fbf: As `recentProfiles` already are sort by most recent used profile, I simply add weight on the `SelectorOption` based on the profile index in the `recentProfiles` array.

Fix #8709 d57757c66cad3427eb8e11c88f1731fe5326e8a7: I made a small mistake previously. The weight I added on the `Connect to` selector option were not used in the `onFilterChange` function in `selectorModal` component.

Edit 11/08:
Resolve #3999 695c5ba67067fd7f113e3b50af9266964f255040 : allow users to specify settings on profile group for each profile provider

Edit 12/08
Resolve #7265 f369b140c6a0f9cbbf0d2e477f085189654ebd66 : added a button to reset global default profile settings and to restore inherited settings on a group.

Edit 15/08
Fix #8065 a9c63b53059c842e31b52d1226aec78c6372fcc3: This commit fixes the issue #8065. Resolve the returned Promise in [showProfileSelector](https://github.com/Eugeny/tabby/blob/a9c63b53059c842e31b52d1226aec78c6372fcc3/tabby-core/src/services/profiles.service.ts#L242C5-L242C24) when selector modal is dismissed.

### DIFF
--- a/tabby-core/src/api/profileProvider.ts
+++ b/tabby-core/src/api/profileProvider.ts
@@ -31,6 +31,22 @@ export type PartialProfile<T extends Profile> = Omit<Omit<Omit<{
     }
 }
 
+export interface ProfileGroup {
+    id: string
+    name: string
+    profiles: PartialProfile<Profile>[]
+    defaults: any
+    editable: boolean
+    collapsed: boolean
+}
+
+export type PartialProfileGroup<T extends ProfileGroup> = Omit<Omit<{
+    [K in keyof T]?: T[K]
+}, 'id'>, 'name'> & {
+    id: string
+    name: string
+}
+
 export interface ProfileSettingsComponent<P extends Profile> {
     profile: P
     save?: () => void
--- a/tabby-core/src/api/profileProvider.ts
+++ b/tabby-core/src/api/profileProvider.ts
@@ -37,7 +37,6 @@ export interface ProfileGroup {
     profiles: PartialProfile<Profile>[]
     defaults: any
     editable: boolean
-    collapsed: boolean
 }
 
 export type PartialProfileGroup<T extends ProfileGroup> = Omit<Omit<{
--- a/tabby-core/src/api/profileProvider.ts
+++ b/tabby-core/src/api/profileProvider.ts
@@ -21,6 +21,11 @@ export interface Profile {
     isTemplate: boolean
 }
 
+export interface ProfileDefaults {
+    id: string
+    //[provider]:
+}
+
 export type PartialProfile<T extends Profile> = Omit<Omit<Omit<{
     [K in keyof T]?: T[K]
 }, 'options'>, 'type'>, 'name'> & {
--- a/tabby-core/src/api/profileProvider.ts
+++ b/tabby-core/src/api/profileProvider.ts
@@ -74,14 +74,12 @@ export abstract class ProfileProvider<P extends Profile> {
     deleteProfile (profile: P): void { }
 }
 
-export abstract class ConnectableProfileProvider<P extends ConnectableProfile> extends ProfileProvider<P> {
+export abstract class ConnectableProfileProvider<P extends ConnectableProfile> extends ProfileProvider<P> {}
 
-    quickConnect (query: string): PartialProfile<P>|null {
-        return null
-    }
+export abstract class QuickConnectProfileProvider<P extends ConnectableProfile> extends ConnectableProfileProvider<P> {
 
-    intoQuickConnectString (profile: P): string|null {
-        return null
-    }
+    abstract quickConnect (query: string): PartialProfile<P>|null
+
+    abstract intoQuickConnectString (profile: P): string|null
 
 }
--- a/tabby-core/src/api/profileProvider.ts
+++ b/tabby-core/src/api/profileProvider.ts
@@ -21,11 +21,6 @@ export interface Profile {
     isTemplate: boolean
 }
 
-export interface ProfileDefaults {
-    id: string
-    //[provider]:
-}
-
 export type PartialProfile<T extends Profile> = Omit<Omit<Omit<{
     [K in keyof T]?: T[K]
 }, 'options'>, 'type'>, 'name'> & {
--- a/tabby-core/src/api/profileProvider.ts
+++ b/tabby-core/src/api/profileProvider.ts
@@ -21,6 +21,10 @@ export interface Profile {
     isTemplate: boolean
 }
 
+export interface ConnectableProfile extends Profile {
+    clearServiceMessagesOnConnect: boolean
+}
+
 export type PartialProfile<T extends Profile> = Omit<Omit<Omit<{
     [K in keyof T]?: T[K]
 }, 'options'>, 'type'>, 'name'> & {
@@ -54,7 +58,6 @@ export interface ProfileSettingsComponent<P extends Profile> {
 export abstract class ProfileProvider<P extends Profile> {
     id: string
     name: string
-    supportsQuickConnect = false
     settingsComponent?: new (...args: any[]) => ProfileSettingsComponent<P>
     configDefaults = {}
 
@@ -68,6 +71,11 @@ export abstract class ProfileProvider<P extends Profile> {
 
     abstract getDescription (profile: PartialProfile<P>): string
 
+    deleteProfile (profile: P): void { }
+}
+
+export abstract class ConnectableProfileProvider<P extends ConnectableProfile> extends ProfileProvider<P> {
+
     quickConnect (query: string): PartialProfile<P>|null {
         return null
     }
@@ -76,5 +84,4 @@ export abstract class ProfileProvider<P extends Profile> {
         return null
     }
 
-    deleteProfile (profile: P): void { }
 }
--- a/tabby-core/src/services/commands.service.ts
+++ b/tabby-core/src/services/commands.service.ts
@@ -109,6 +109,6 @@ export class CommandService {
                 description: c.sublabel,
                 icon: c.icon,
             })),
-        )
+        ).catch(() => {return})
     }
 }
--- a/tabby-core/src/services/commands.service.ts
+++ b/tabby-core/src/services/commands.service.ts
@@ -101,7 +101,7 @@ export class CommandService {
             context.tab = tab.getFocusedTab() ?? undefined
         }
         const commands = await this.getCommands(context)
-        await this.selector.show(
+        return this.selector.show(
             this.translate.instant('Commands'),
             commands.map(c => ({
                 name: c.label,
@@ -109,6 +109,6 @@ export class CommandService {
                 description: c.sublabel,
                 icon: c.icon,
             })),
-        ).catch(() => {return})
+        ).then(() => {return})
     }
 }
--- a/tabby-core/src/services/fileProviders.service.ts
+++ b/tabby-core/src/services/fileProviders.service.ts
@@ -13,8 +13,9 @@ export class FileProvidersService {
     ) { }
 
     async selectAndStoreFile (description: string): Promise<string> {
-        const p = await this.selectProvider()
-        return p.selectAndStoreFile(description)
+        return this.selectProvider().then(p => {
+            return p.selectAndStoreFile(description)
+        })
     }
 
     async retrieveFile (key: string): Promise<Buffer> {
--- a/tabby-core/src/services/profiles.service.ts
+++ b/tabby-core/src/services/profiles.service.ts
@@ -214,7 +214,7 @@ export class ProfilesService {
     }
 
     getConfigProxyForProfile <T extends Profile> (profile: PartialProfile<T>, skipUserDefaults = false): T {
-        const defaults = this.getProfileDefaults(profile).reduce(configMerge, {})
+        const defaults = this.getProfileDefaults(profile, skipUserDefaults).reduce(configMerge, {})
         return new ConfigProxy(profile, defaults) as unknown as T
     }
 
--- a/tabby-core/src/services/profiles.service.ts
+++ b/tabby-core/src/services/profiles.service.ts
@@ -2,7 +2,7 @@ import { Injectable, Inject } from '@angular/core'
 import { TranslateService } from '@ngx-translate/core'
 import { NewTabParameters } from './tabs.service'
 import { BaseTabComponent } from '../components/baseTab.component'
-import { PartialProfile, Profile, ProfileProvider } from '../api/profileProvider'
+import { PartialProfile, Profile, ProfileDefaults, ProfileProvider } from '../api/profileProvider'
 import { SelectorOption } from '../api/selector'
 import { AppService } from './app.service'
 import { configMerge, ConfigProxy, ConfigService } from './config.service'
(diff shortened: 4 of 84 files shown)


Eugeny:
LGTM so far :+1:

Clem-Fern:
Hey @Eugeny,

I've done a bit of refactoring on Profile & ProfileGroup system, I tried to move all the interaction (creation, update, deletion) with Profile and ProfileGroup in ProfileService as discuss in #8663. My goal is to make maintenance and future work easier by having only one entrypoint for profile operation. With the methods in the core service, It will allow me to remove direct access to `config.store`  from `profilesSettingsTab.component.ts`.

I also tried to prepare implementation of #3999, #3880, #7265 and #6650 by reworking ProfileGroup. Should I create a dedicated service for Group ? (ProfileGroupService for example)

Could you, please, take a look when you got the time ? Even if I'm far from done, I would reallly like to have your feedback on this :)
You will probably have a better approach than mine to suggest

Eugeny:
Wow! I haven't had time to review it in detail yet, but I like your direction. 

Small nitpick: I wouldn't keep the collapsed status in the config, it's only relevant for the profileSettingsTab anyway. 
What's the reason for `writeProfile` deleting the fields one by one? Is it so that default values get removed? That should already work automatically via ConfigProxy.

One thing to keep in mind (now that you have 3 config "levels") is what's going to happen if a profile setting is changed to non-default in the group config, but then is changed back to default in one of its profiles. Right now ConfigProxy will remove the value completely since it matches the defaults. You'll need to make sure that `getProfileDefaults` includes group config as the profile's "defaults".

Clem-Fern:
> Wow! I haven't had time to review it in detail yet, but I like your direction.

Glad to hear that ! ;)



> Small nitpick: I wouldn't keep the collapsed status in the config, it's only relevant for the profileSettingsTab anyway.

Oh yeah, seems to be the good way now you say it.



> One thing to keep in mind (now that you have 3 config "levels") is what's going to happen if a profile setting is changed to non-default in the group config, but then is changed back to default in one of its profiles. Right now ConfigProxy will remove the value completely since it matches the defaults. You'll need to make sure that `getProfileDefaults` includes group config as the profile's "defaults".

I haven't really looked into it yet but it was my next step planned.

Clem-Fern:
> What's the reason for `writeProfile` deleting the fields one by one? Is it so that default values get removed? That should already work automatically via ConfigProxy.

When a profile is updated, the group attribute is set to `undefinied` if no group has been selected. Because of that, `Object.assign` do not seem to override the group attribute and leave the old value. It made impossible to reassign a profile in `Ungrouped`. But, now at head rested, I think this behavior only happens with group. It would probably be easier to change a bit the EditProfileModal to leave an empty string in the group attribute when no group has been selected.

Clem-Fern:
Hi @Eugeny, I hope you are doing great :)

Resolve #3999 695c5ba67067fd7f113e3b50af9266964f255040 : With this commit, I got a first usable version allowing users to specify settings on profile group for each profile provider. I will use this branch as my daily to dig up any bug.

Clem-Fern:
Resolve #7265 f369b140c6a0f9cbbf0d2e477f085189654ebd66 : I added a button to reset global default profile settings and to restore inherited settings on a group. I'm not sure that I correctly understand what was asked in #7265 but I think this will still be a great feature.
I also wander if I should add a tooltip on the button to describe what it does and if it would not be useful to add a second button on groups to set the settings with the app defaults. What's your point on this ?

image

Clem-Fern:
No problem, there's absolutely no rush. I got two more issues I would like to address so I think I will stop adding new commit soon anyway. It's a good first step for me I think and really don't want to make this PR a chore to review for you :)

Clem-Fern:
ef6b8a4eaa1ce4cf0311649d009d1a8a3e66a71c : I made a bit of refactoring to be aligned with my previous PR #8416. I created an interface named `ConnectableProfile` which inherit the `Profile` interface attributes. I was thinking that it could be useful to be able to declare properties on connectable profile only (in anticipation of ef6b8a4eaa1ce4cf0311649d009d1a8a3e66a71c). To go with this change, I also create an abstract class `ConnectableProviderProfile` and moved `quickConnect` and `intoQuickConnectString` methods in it as the quick connect capability only concerns connectable profiles. I updated serial/ssh/telnet session, tab component and provider class to concur with this refactoring.

ef6b8a4eaa1ce4cf0311649d009d1a8a3e66a71c : I added a new property in `ConnectableProfile` named `clearServiceMessagesOnConnect`. My goal was to replace the `clearServiceMessagesOnConnect` specific ssh option and provide a per connectable provider type option. The `clearServiceMessagesOnConnect` attribute is now read in the `initializeSession` method and the terminal cleared according to the attribute value. As this feature did not exists before for the telnet and serial session, the default value was setted to false.

5eeaef954c8f5f3e8a60c98a8b055a718d16423a : I removed the old `clearServiceMessagesOnConnect` option and added a new config migration to suit the previous commits. The old clearServiceMessagesOnConnect property is transferred into the ssh provider profile defaults.

Clem-Fern:
Fix #8065 a9c63b53059c842e31b52d1226aec78c6372fcc3: This commit fixes the issue #8065. While trying to understand why some button became unresponsive when `Profiles & Connections` selector is closed without selecting anything, I found that dismissing the selector modal was not handle in [showProfileSelector](https://github.com/Eugeny/tabby/blob/a9c63b53059c842e31b52d1226aec78c6372fcc3/tabby-core/src/services/profiles.service.ts#L242C5-L242C24). It causes the returned promise to not being resolved or rejected.

Even if unhandled rejection on modal dismissed does not seem to have impact anywhere else than for #8065, I also tried to add a `catch` call to handle Promise rejection. It avoids raising `Uncaught (in promise)` exception and made the log a bit cleaner.

tabby-core/src/services/fileProviders.service.ts
```diff
@@ -13,8 +13,9 @@ export class FileProvidersService {
     ) { }
 
     async selectAndStoreFile (description: string): Promise<string> {
-        const p = await this.selectProvider()
-        return p.selectAndStoreFile(description)
+        return this.selectProvider().then(p => {
```
Eugeny, line 16: no behaviour changes here?

tabby-core/src/services/profiles.service.ts
```diff
@@ -36,6 +39,145 @@ export class ProfilesService {
...
+            return this.config.save()
+        }
+    }
+
+    /**
+    * Delete all Profiles from config using option filter
+    * arg: options { group: string } -> options used to filter which profile have to be deleted
+    * arg: saveConfig (default: true) -> invoke after the Profile was deleted
+    */
+    async deleteBulkProfiles (options: { group: string }, saveConfig = true): Promise<void> {
```
Eugeny, line 161: would be more future proof with a predicate function parameter instead of a filter

tabby-core/src/services/profiles.service.ts
```diff
@@ -177,30 +307,38 @@ export class ProfilesService {
...
+                            callback: query => {
+                                const profile = provider.quickConnect(query)
+                                resolve(profile)
+                            },
+                        })
+                    }
                 })
 
-                await this.selector.show(this.translate.instant('Select profile or enter an address'), options)
+                await this.selector.show(this.translate.instant('Select profile or enter an address'), options).catch(() => resolve(null))
```
Eugeny, line 326: same (swallows modal dismissal)

tabby-core/src/services/profiles.service.ts
```diff
@@ -211,27 +349,192 @@ export class ProfilesService {
...
+    setProviderDefaults (provider: ProfileProvider<Profile>, pdefaults: any): void {
+        this.config.store.profileDefaults[provider.id] = pdefaults
+    }
+
+    /**
+    * Return defaults for a given profile
+    * Always return something, empty object if no defaults found
+    * arg: skipUserDefaults -> do not merge global provider defaults in ConfigProxy
+    */
+    getProfileDefaults (profile: PartialProfile<Profile>, skipGlobalDefaults = false, skipGroupDefaults = false): any[] {
```
Eugeny, line 378: since this is a public api, it needs to be future proof by using an optional `options` object arg instead of separate params

tabby-core/src/services/profiles.service.ts
```diff
@@ -211,27 +349,192 @@ export class ProfilesService {
...
+    getSyncProfileGroups (): PartialProfileGroup<ProfileGroup>[] {
+        return deepClone(this.config.store.groups ?? [])
+    }
+
+    /**
+    * Return an Array of the existing ProfileGroups
+    * arg: includeProfiles (default: false) -> if false, does not fill up the profiles field of ProfileGroup
+    * arg: includeNonUserGroup (default: false) -> if false, does not add built-in and ungrouped groups
+    */
+    async getProfileGroups (includeProfiles = false, includeNonUserGroup = false): Promise<PartialProfileGroup<ProfileGroup>[]> {
```
Eugeny, line 406: same here (options object)

tabby-core/src/services/profiles.service.ts
```diff
@@ -211,27 +349,192 @@ export class ProfilesService {
...
+
+        return groups
+    }
+
+    /**
+    * Insert a new ProfileGroup in config
+    * arg: saveConfig (default: true) -> invoke after the Profile was updated
+    * arg: genId (default: true) -> generate uuid in before pushing Profile into config
+    */
+    async newProfileGroup (group: PartialProfileGroup<ProfileGroup>, saveConfig = true, genId = true): Promise<void> {
```
Eugeny, line 470: same here

tabby-core/src/services/profiles.service.ts
```diff
@@ -211,27 +349,192 @@ export class ProfilesService {
...
+                return this.config.save()
+            }
+        }
+    }
+
+    /**
+    * Delete a ProfileGroup from config
+    * arg: saveConfig (default: true) -> invoke after the ProfileGroup was deleted
+    */
+    async deleteProfileGroup (group: PartialProfileGroup<ProfileGroup>, saveConfig = true, deleteProfiles = true): Promise<void> {
```
Eugeny, line 509: please avoid having a saveConfig argument in every function, that's the caller's responsibility

tabby-core/src/api/profileProvider.ts
```diff
@@ -31,6 +35,21 @@ export type PartialProfile<T extends Profile> = Omit<Omit<Omit<{
...
 
+export interface ProfileGroup {
+    id: string
+    name: string
+    profiles: PartialProfile<Profile>[]
+    defaults: any
+    editable: boolean
+}
+
+export type PartialProfileGroup<T extends ProfileGroup> = Omit<Omit<{
```
Eugeny, line 46: what does this type do?

Eugeny (changes requested):

Eugeny:
I've gone through it and left some nitpicks here and there but all in all the PR looks great already :+1:

tabby-core/src/services/fileProviders.service.ts
Clem-Fern: It should not, I didn't notice any but my understanding could be wrong.
I tried to avoid `Uncaught error in Promise` cause by awaiting `selectProvider`. Chaining promise here allow me to return a rejected promise if `selectProvider` is rejected too.

tabby-core/src/services/commands.service.ts
```diff
@@ -109,6 +109,6 @@ export class CommandService {
                 description: c.sublabel,
                 icon: c.icon,
             })),
-        )
+        ).catch(() => {return})
```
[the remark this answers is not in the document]
Clem-Fern: Yeah you're right, I should have place the catch call on a higher side. Same on the other you've point to me.

Corrected with 49f9a10372d15f7e6b358ba9a43536a8e188e8b7 :
In that case, `showSelector` now return the promise return by the selector. I replaced `catch` by `then` because of the type of the promise (`Promise<void>`).

tabby-core/src/services/profiles.service.ts
Clem-Fern: Corrected with 49f9a10372d15f7e6b358ba9a43536a8e188e8b7 :
It was a bit dumb of me to resolve the Promise when the selector is dismissed. The Promise is now rejected on selector dismiss.

tabby-core/src/services/profiles.service.ts
Clem-Fern: Resolved in 06859de2de651867be65b0230b679908b58f2fc6

tabby-core/src/services/profiles.service.ts
Clem-Fern: Resolved in 5e5c80832de2d2125a9e55d89321071365c32938

tabby-core/src/services/profiles.service.ts
Clem-Fern: Resolved in 5e5c80832de2d2125a9e55d89321071365c32938

tabby-core/src/services/profiles.service.ts
Clem-Fern: Resolved in 5e5c80832de2d2125a9e55d89321071365c32938

tabby-core/src/services/profiles.service.ts
Clem-Fern: Resolved in 0128013308bbb307275b559b4ce6984581785ebd:
I'm not sure that I've done it the good way. Hope It's okay.

tabby-core/src/api/profileProvider.ts
Eugeny: nevermind just took me 24 hours to parse it in my head

Eugeny (approved):

Eugeny:
Phew what a huge merge 😅

Clem-Fern:
Hey, I'm sorry, I didn't have the time to respond and fix everything in one time...
Thank you for taking the time to review this and for finishing cleaning my mess ahah

Refactoring profiles service was the first step to achieve before being able to work on profiles/groups tree hierarchy. I would like to work on this next (if you agree ofc).

Eugeny:
Thanks for undertaking the work haha

I'm fine with nested groups as long as it's not negatively affecting the "normal" case that 90% of users will have (no nested groups), e.g. so that the profile selector doesn't end up being some unwieldy MobaXTerm-esque control

PR #8801, opened 2023-08-07
perf: avoid gpu memory increase

robinfai:
This modification can effectively control the memory usage growth of GPU processes as the number of tabs increases. In our internal terminal client, this modification achieved a 75% reduction in overall memory usage at the P95 percentile.


Eugeny:
Sorry for the delay reviewing your PR - everything looks good, but I've added a delay for canvas downscaling to avoid flickering when quickly switching tabs

Issue #8912, opened 2023-09-01
1.0.200 - clear SSH service messages by default (not working)


vecinohk:
Hi Eugene, this feature stopped working after the version update to 1.0.200.

https://github.com/Eugeny/tabby/issues/7841
https://github.com/Eugeny/tabby/commit/7491c3119f3a7911341e33c98684213b0dddd6fa

Thank you

```
Version: 1.0.200
Platform: Windows x64 10.0.22621
Plugins: none
Frontend: xterm-webgl
```
btw: And thanks a lot for the Czech translation.

Clem-Fern:
Hey,
I apologize, definitely my bad. Bug related to #8726 ef6b8a4eaa1ce4cf0311649d009d1a8a3e66a71c
I'm opening a new PR address this issue.

vecinohk:
Hey Clem,
never mind it will happen - there are worse things in the world after all. ✌️ 

Thanks and good luck!

Clem-Fern:
You can try this build : https://github.com/Eugeny/tabby/actions/runs/6059256578
Service message should now be clear after connection.

vecinohk:
For some profiles this is already fine, but for some even if I turn on the option "clear terminal after connection" and click save and go back to the profile, the switch is still disabled and the function does not work.

`Version: 1.0.201-nightly.0`

Clem-Fern:
That's weird, do you have any error in debug console ?
https://github.com/Eugeny/tabby/wiki/Debug-logs

Just to be sure, if you tried with profile already opened in tab, you have to close the tab and open it again for clear terminal option to take effect.

vecinohk:
Yes, I have repeatedly quit and restarted the entire application. In the console there are the following errors/warnings.

```
The vm module of Node.js is deprecated in the renderer process and will be removed.

[Intervention] Slow network is detected. See https://www.chromestatus.com/feature/5636954674692096 for more details. Fallback font will be used while loading: file:///E:/Software/Tabby/resources/app.asar/dist/ac3c0d6660d55a57bacf.woff2
10DevTools failed to load source map: Could not load content for file:///E:/Software/Tabby/app.asar.unpacked/assets/webpack/index.js.map: Unexpected end of JSON input
DevTools failed to load source map: Could not load content for file:///E:/Software/Tabby/resources/app.asar/dist/index.js.map: Unexpected end of JSON input
[...]

PR #8918, opened 2023-09-02
Fix#8912: clear SSH service messages not working

Clem-Fern:
Hey @Eugeny, I hope you are doing well.

This PR aims to fix the issue #8912. By adding a global connectable tab option for `clearServiceMessagesOnConnect` in #8726 ef6b8a4eaa1ce4cf0311649d009d1a8a3e66a71c, I placed the fronted clear call before session finished to initialize. It results in only cleaning latest terminal session log on reconnection and not service message on connection, which is not intended.

Feel free to ask if any change are needed.


Eugeny:
LGTM 🚀

PR #8926, opened 2023-09-04
Fix#8925: unable to set back a value to the inherited default value on profile

### LINKED ISSUE #8925: Profile Settings: unable to set back an option to the inherited default value on profile
Version: 1.0.198 -> 1.0.200

**Describe the problem**:
Unable to set back a value on profile when it corresponds to the inherited default value.
It seems to affect only `Disable dynamic tab title` and `Clear terminal after connection`.

Issue reported by @vecinohk in #8912

**To Reproduce**:
1. Enable `Clear terminal after connection` in `Profiles & connections > Advenced > Default profile settings` for Telnet (or SSH/Serial)
2. Create a new Telnet profile, disable `Clear terminal after connection`, then save the profile.
3. Edit the newly created profile and enable `Clear terminal after connection`, then save.
4. You will notice that you can't enable again `Clear terminal after connection` option on that profile.

Clem-Fern:
Hey @Eugeny,

As discuss in #8912 and describe in #8925, my previous PR #8726 seems to have bring a new bug regarding the `Disable dynamic tab title` and `Clear terminal after connection` options (really sorry about that :/). At the time of the PR, I only notice this problem with the `group` attribute. I think we do not have other choice than deleting profile fields one by one in `writeProfile` method to prevent having ghost attributes in the profile config. 

Without deleting the field one by one, `Object.assign` call does not delete `undefinied` value from the profile object (Context from #8726):
> > What's the reason for `writeProfile` deleting the fields one by one? Is it so that default values get removed? That should already work automatically via ConfigProxy.

> When a profile is updated, the group attribute is set to `undefinied` if no group has been selected. Because of that, `Object.assign` do not seem to override the group attribute and leave the old value. It made impossible to reassign a profile in `Ungrouped`. But, now at head rested, I think this behavior only happens with the group attribute. It would probably be easier to change a bit the EditProfileModal to leave an empty string in the group attribute when no group has been selected.

Anyway, as we now provide profile object copy with `getProfileGroups`, profile field suppression can only be done inside `writeProfile` method. (related refactoring in e4169a54aa1686e928e81fd1391351304815407a)

I tried my best to describe what I found, I hope all of this is clear enough ^^'
Feel free to ask if any changes needed

Resolve #8925

### DIFF
--- a/tabby-core/src/services/profiles.service.ts
+++ b/tabby-core/src/services/profiles.service.ts
@@ -115,10 +115,11 @@ export class ProfilesService {
     async writeProfile (profile: PartialProfile<Profile>): Promise<void> {
         const cProfile = this.config.store.profiles.find(p => p.id === profile.id)
         if (cProfile) {
-            if (!profile.group) {
-                delete cProfile.group
+            // Fully replace the config
+            for (const k in cProfile) {
+                // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+                delete cProfile[k]
             }
-
             Object.assign(cProfile, profile)
         }
     }
--- a/tabby-settings/src/components/profilesSettingsTab.component.ts
+++ b/tabby-settings/src/components/profilesSettingsTab.component.ts
@@ -70,25 +70,24 @@ export class ProfilesSettingsTabComponent extends BaseComponent {
                 return
             }
         }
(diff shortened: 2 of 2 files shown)


Eugeny:
Makes sense and thank you for the fix!

Issue #4327, opened 2021-08-02
[Feature Request] Custom configuration file path


dallaslu:
The default path is `tabby/config.yaml`, there are many other files in the same folder. This makes it difficult to sync using a web drive client such as Dropbox, Nextcloud.

Once you can customize the path, it's easy to switch configurations (may be useful in some very specific cases).

At the very least, key text files such as config.yaml can be moved to a folder. For Example: `tabby/con/config.yaml`.

clarkwang:
What about adding something like an `includes` option? Then it'll also be easier to share the conf between Mac/Windows/Linux.

~~~
includes:
  - /my/tabby/common.yaml
  - /my/tabby/win10.yaml
~~~

dallaslu:
What about the changes made in the tabby settings screen, are they saved to config.yaml or the include file?

Issue #6842, opened 2022-07-26
custom config file location


nimdasx:
**Is your feature request related to a problem? Please describe.**
no, is not related to problem

**Describe the solution you'd like**
i would like using custom config file location, without third party plugin like config sync

thanks for this great software

PR #9048, opened 2023-09-30
added possibility to set config file path in environment file

nwalser:
Hi Eugeny

This PR adds the ability to set the config file via an environment variable. The variable can be set in a .env file.
For now you would need to edit the .env file manually to move the config file to another location.

Is there a particular problem with the current implementation? Or is it possible to continue on this path and add the needed UI changes?

Resolves: #4327 #6842


Eugeny:
Hi, thanks for the contribution! This looks good to me so far, though I would rename the var to `TABBY_CONFIG_FILE_PATH` since your name is generic enough that it could have been set on a user's system by something else.

Eugeny:
Thank you! @all-contributors add @nwalser for code

3lorsin:
Hi @Eugeny and @nwalser 

I am having a difficult time finding where the .env file is, or where I need to add it to my filesystem. I have both a Windows 10 machine (installed via the setup exe) and a Debian Linux machine (installed via packagecloud).

Maybe this info could be added to the Repo's Wiki Config-File page?

Thank You!

nwalser:
Hi @3lorsin
This would be quite helpful indeed. I will have a look at it in a few days as I am currently busy.
I'll keep you posted. Cheers, Nathaniel

nwalser:
@3lorsin I will answer your question here as I am not able to edit the wiki pages currently.

Tabby will read out the value of the TABBY_CONFIG_DIRECTORY environment variable, in which you can define a directory path to which the config file will be stored/loaded. You have two options to set the desired path:
- You can set this environment variable on the operating system level according to your specific operating system.
- You can place a text file named `.env` inside the installation directory of tabby (e. g. on windows "C:\Users\{username}\AppData\Local\Programs\Tabby") with the variables as key value pairs. (`TABBY_CONFIG_DIRECTORY="{PATH_TO_DIRECTORY}"`. Tabby will read these key value pairs and inject them into the application as if they where environment variables set by the operating system.

> [!NOTE]
> If you have a current configuration for tabby already, you will have to move this file manually to the new place. If there is no config file in the new place an empty one gets created for you.

Let me know if this worked for you, or if I may clarify something for you. :)

3lorsin:
@nwalser This is working perfectly for me. Thank You for the information and adding this feature!

i-api:
for some reason on MacOS setting TABBY_CONFIG_DIRECTORY does not work, can we reopen this issue?

PR #9232, opened 2023-11-14
Fix: native modules rebuild on macos

Clem-Fern:
Hi @Eugeny, I hope you're doing fine.

As describe in actions/runner#2958, node-gyp rebuild failed because of a missing dependency.
Adding `setuptools` pip module before building seems to fix this issue (https://github.com/Clem-Fern/tabby/actions/runs/6868200937).


Eugeny:
Nicely done 👌

PR #9234, opened 2023-11-15
add tabby-background plugin.

moemoechu:
I made a plugin to change tabby background image and UI font.
Hope everyone may glade to use it.


Eugeny:
Nice job!

PR #9247, opened 2023-11-18
Add tabby-highlight plugin.

moemoechu:
Add my tabby-highlight plugin.
Hope to bring more convenience to everyone.


Jackz9949:
Hello, now this plugin is not available for download via the tabby plugin

PR #9238, opened 2023-11-15
Fix builds

Clem-Fern:
Hey @Eugeny,

As describe in #9235, latest linux release is unusable. Sadly, I couldn't identify precisly where the problem was coming from.
I think it was related to node-gyp 10.0.0 dependecy in the root project. 0f6561e85922f38700cd328747a0ccacf4644274

I do not have any mac device with me. Could you try this build ? https://github.com/Eugeny/tabby/actions/runs/6883409733

There is also a weird behavior on the macos-12 runner image. The default python version for macos arm64 is python2 (python 3 on x86_64).
it does not seem to be an issue right now but it has to be kept in mind as this python version is now deprecated. It could bring some difficulty in the future.

Fix #9235
Fix #9242
Fix #9245

PR #9249, opened 2023-11-18
fix: force display title-bar when tab-bar on left/right

Clem-Fern:
Hey @Eugeny,

This PR changes a bit the title bar behavior when :
- Window frame = `Thin`
- Tabs location = `Left` or  `Right`

Before:
image

With this PR:
image

It is actually the same behavior as Window frame `Full` but I find it a lot more user friendly and practical that way.
What do you think ?


Eugeny:
Screenshot 2023-11-18 at 21 24 09

Can you enable this only on Linux & Windows? On macOS, tabby has special provisions for window controls on the left side

Clem-Fern:
Arf... definitely failed my merge sorry...
We can wait for #9238 to be ok before merging this. Or I can recreate a proper PR.

Anyway, good to know for macos thank's, the platform should not be impacted now 191807c09aec410d38add65b9e94c510000cf45a

PR #9251, opened 2023-11-19
Win Portable: fix data directory and updater issue

### LINKED ISSUE #7768: Portable Windows updates as installation, not as portable
Portable Windows Tabby with "Check for updates" gets a new version (like from 1.0.187 to 1.0.188), but it runs installation, asking if it should be installed for a user or for all users. 
It should instead just extract to the same folder of Tabby.exe so that it remains portable.

### LINKED ISSUE #9148: The portable version configuration file cannot be placed in the software directory on the C drive
How to put the configuration file into the software directory in the portable version? Only in this way can it be truly portable and convenient to put on a USB flash drive for portable use.

### LINKED ISSUE #9243: Update 1.0.200 > 1.0.204 - Lost Data / Profils
**Describe the problem**:
I use Tabby version 1.0.200 portable ( tabby-1.0.200-portable-x64.zip  ) 
image
image

#### After Update 1.0.200 to 1.0.201 :
image
image

Lost Profils and Data
The vault remembers my password, but it's empty too.

1.0.200 to 1.0.204 same issue

**To Reproduce**:
Download tabby-1.0.200-portable-x64.zip and create Profils, ...
Update to tabby-1.0.201-portable-x64.zip or tabby-1.0.204-portable-x64.zip

Lost Profils

Clem-Fern:
Hey @Eugeny,

2be079a51b6c668a217bdd7f77886f63e786763a: The new `TABBY_CONFIG_DIRECTORY` env var added in #6242 bypassed `data` folder on portable version.

1e096ede7785eca00da99eff7e96ab68590805ad: I added an exclusion for portable version in the updater service. Otherwise, the electron installer is used when an update is available.

Fix #9243
Fix #9148
Fix #7768

### DIFF
--- a/app/lib/index.ts
+++ b/app/lib/index.ts
@@ -1,5 +1,8 @@
 import { app, ipcMain, Menu, dialog } from 'electron'
 
+// set userData Path on portable version
+import './portable'
+
 // set defaults of environment variables
 import 'dotenv/config'
 process.env.TABBY_PLUGINS ??= ''
@@ -7,7 +10,6 @@ process.env.TABBY_CONFIG_DIRECTORY ??= app.getPath('userData')
 
 
 import 'v8-compile-cache'
-import './portable'
 import 'source-map-support/register'
 import './sentry'
 import './lru'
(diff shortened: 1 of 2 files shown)


Eugeny:
Nicely done 👌

PR #9253, opened 2023-11-19
fixed keyboard shortcut (control + backspace) in windows

zKXDEX:
I was trying some keyboard shortcuts and so far this was the only one that didn't work in windows.


Eugeny:
Thank you for the fix! @all-contributors please add @zKXDEX  for code

ghostrider-reborn:
https://github.com/Eugeny/tabby/issues/9276 seems to be caused by this. i would suggest reverting or making it conditional.

AY1uZwIcJzKhaGywovQP:
This commit causes the `delete-previous-word` shortcut to send `Ctrl + W` which in bash shell causes it to delete including semi-colon, hyphen, and the like (unlike the current one which treats them as word separators). Would it be possible for it to send `Alt + Backspace` for `delete-previous-word` (or assign that signal to a new hotkey)?

PR #9307, opened 2023-12-06
add implicit CR and implicit LF for serial terminal similar to putty

kimbob13:
Hello,
Firstly, I've tried several terminal programs for serial connection, and I've found the tabby is really satisfying experience for me,
so thanks for the great development.

The one thing I missed from other programs is the "**Implicit CR in every LF**" and "**Implicit LF in every CR** features, which can be found in like putty.
That is, the tabby has a feature for "Force CRLF", but I want to add a CR only in LF, not adding LF in CR.

Therefore, this PR adds following two setting entries for new line modes in serial connection.
- Imiplicit CR: Only add CR in LF
- Implicit LF: Only add LF in CR

### DIFF
--- a/locale/ko-KR.po
+++ b/locale/ko-KR.po
@@ -916,6 +916,14 @@ msgstr "CRLF 강제"
 msgid "Force LF"
 msgstr "LF 강제"
 
+#: tabby-terminal/src/components/streamProcessingSettings.component.ts:56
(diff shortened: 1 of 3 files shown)


Eugeny:
Thank you for the contribution! @all-contributors please add @kimbob13 for code

PR #8613, opened 2023-06-24
Cross Compile linux armhf/arm64 artifacts

Jai-JAP:
@Eugeny
This (finally!!) implements a better way to compile tabby for arm linux platforms than docker, quite complex than before but faster. 
Completes in ~25-35 mins as compared to ~2 hours with docker based compilation method used previously.

EDIT: Also added AppImage target for linux

Previous approach is still left **_commented_** in place. 
***TESTED working on both arm64 and armhf***

Also updated dependency from `@tabby-gang/node-pty` to `note-pty` as it has a new beta release in March. (Feel free to discard this.)

Add better cross-compilation support to `scripts/vars.mjs`

Prefer squash merge on this PR due to its dirty commit history.

PR #9355, opened 2023-12-23
Better crossbuild v3 + fixes

Jai-JAP:
This PR implements a slightly better and a faster way to cross-build arm(64) artifacts.

The previous approach uses a x64 bionic chroot to build the native modules, whereas this approach uses a bare bionic arm(64) debootstrap as a sysroot which is just used to link arm(64) libraries using `--sysroot` parameter passed to [gcc/g++](https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html#index-sysroot) and `PKG_CONFIG_PATH` to specify path of `*.pc` files to [pkg-config](https://people.freedesktop.org/~dbn/pkg-config-guide.html).

Thus this approach on its own should give a slight improvement as compared to a bionic chroot plus the added benefit of caching the sysroot makes it even faster than the previous approach implemented in #8613

Build times have been reduced from 15-20 mins to 11-13 mins.

Additionally this PR moves to Node 18 as version 16 now EOL, removes old code & fixes a potential ambiguous case in scripts/vars.mjs.

This PR is not tested but it should not break anything theoretically.


Eugeny:
Thanks! Have you tested the produced ARM builds? I can test the rest

Jai-JAP:
> Thanks! Have you tested the produced ARM builds? I can test the rest

Nope, my pi is no longer working so can't test it.

SmollClover:
Hi there! Is testing still required for this? I happen to have a working Raspberry Pi 400 at home and I could test the build if need be. I think it would be best to update this PR with the latest commits from the master branch before I do that though, to make sure that the latest release works and not just an old one.

The question would then also be what parts I should test exactly. Since there isn't a test suite or a written list of functions to specifically test for Tabby, I would love to hear what is roughly always tested before each Release is made.

Eugeny:
@SmollClover thank you! It should be enough to just do basic testing of opening tabs & connecting to some SSH server as that will cover the most arch-related things (native Node modules in use)

SmollClover:
Okay, I've tested the following things now:

- Normal Terminal
- SSH Connection
  - Password Authentication
  - Certificate Authentication
- Multiple Tabs
  - Having them open
  - Moving them around
  - Coloring them
- SSH Port Forwarding
- Downloading and Uploading Files through SFTP
- Using a theme
- Installing Plugins
- Syncing the Config
- Using the Vault
  - Passwords
  - Certificates
  - Passwords for the Certificates
- Encrypt config file

All of the above I have checked and can verify that those functions work without any issues exactly how it currently works on the latest release. I took the liberty of including most of the features I use on a daily basis in Tabby and I haven't found an error or inconsistency as to how they normally work. 👍🏻

Eugeny:
Thank you! 🤝

Issue #9354, opened 2023-12-23
Unable to resolve ipv6 address properly when right clicking to open WinSCP session while using IPV6 to connect to SSH


Mxmilu666:
**Describe the problem**:
Unable to resolve ipv6 address properly when right clicking to open WinSCP session while using IPV6 to connect to SSH

**To Reproduce**:
Right click to open WinSCP session after SSH connection using IPV6 address

PR #9359, opened 2023-12-24
Fix Connection error while opening WinSCP session using IPV6 address

Mxmilu666:
Fix Connection error while opening WinSCP session using IPV6 address


Mxmilu666:
#9354

PR #9369, opened 2023-12-29
Add web-auth-handler to plugin list

BowsiePup:
This plugin aims to make using web authentication easier by intercepting the interactive keyboard prompt and opening the url in a popup window as a child to the main app 

Screenshot:
image


Eugeny:
Looking great, thanks!

PR #9383, opened 2024-01-11
Add Support for a Quick Connect CLI command

ashneilson:
The goal of this PR is to introduce a new CLI command that enables launching an `ssh` or `telnet` connection in a new tab using the Profile Quick Connect query format.

Examples of utilizing the new command:

#### SSH
```bash
tabby quickConnect ssh 'bob@ssh-hostname.com:22'
```

#### Telnet
```bash
tabby quickConnect telnet '192.168.0.123:9000'
```

### DIFF
--- a/app/lib/cli.ts
+++ b/app/lib/cli.ts
@@ -28,6 +28,16 @@ export function parseArgs (argv: string[], cwd: string): any {
         .command('recent [index]', 'open a tab with a recent profile', {
             profileNumber: { type: 'number' },
         })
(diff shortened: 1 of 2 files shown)


Eugeny:
This looks great - thank you for the contribution! @all-contributors please add @ashneilson for code

thschaffr:
@Eugeny, I would like to use Tabby with telnet links from the browser. (For CLI access)
Do I have to configure certain parameters etc. in the application?
Right now Tabby launches but doesn't create the telnet connection. 
Thank you.

PR #9410, opened 2024-01-23
Improve sftp visual effects

cfs4819:
### Improve sftp visual effects

image

1. Automatically select the corresponding font-awsome icon based on the file suffix.

2. Set the border effect of hover


cfs4819:
The border effect is shown as follow.

hover

PR #9463, opened 2024-02-19
chore: changed grammar in de-DE.po

0xflotus:


Eugeny:

PR #9464, opened 2024-02-19
chore: added translation to de-DE.po

0xflotus:


Eugeny:

Issue #8953, opened 2023-09-08
Always print “Host key fingerprint:”，cannot cancel


panhongyin:
Version: 1.0.200
Platform: Windows x64 10.0.22621
Plugins: search-in-browser
Frontend: xterm-webgl

![image](https://github.com/Eugeny/tabby/assets/23182057/f5e77c98-9130-47a4-ac00-c06e477e59ef)

dzqs1993:
+1

Horizon2333:
Any solution?

bifo8:
This seems like something that should be off by default and can be turned on optionally.  It's pointless and irritating for the vast majority of users.

rcasta74:
Hello,
is this issue still present in newer versions?
I'm using this program since few days but I'm stuck with version 1.0.215 due to agent issue.
I don't understand why fingerprint verification is repeated some time, but it is quite annoying when it happens while I'm editing some file with vim.
I don't think this information is usefull for the user, it would be nice if it can be printed only in the log file and displayed only in case of error if the fingerprint does not match with the saved one.

Thanks for this great program.

Issue #8934, opened 2023-09-06
Always print “SSH  Host key fingerprint:”


panhongyin:
When I use ssh to connect to my server, logging in with the key is normal, but logging in with the password will prompt the following message every once in a while:
"
SSH  Host key fingerprint:
SSH   ssh-rsa  JaIr1C364nDL2Lv+UTT7ehpfvMlos76z3Hc2EZLVhik=
"

Version: 1.0.197
Platform: Windows 11
Frontend: xterm-webgl

panhongyin:
@Eugeny  this is a bug in "tabby-ssh/src/session/ssh.ts":
![image](https://github.com/Eugeny/tabby/assets/23182057/08f4ac36-edf0-4832-954c-8ef82d0232f4)

Should be:
![image](https://github.com/Eugeny/tabby/assets/23182057/e84fb3b5-c82b-4772-9034-68387dc58f2b)

cfs4819:
[...]

PR #9544, opened 2024-03-22
Fix: Resolve SSH X11 Forwarding Repeated Handshake and Key Confirmation Issue

cfs4819:
Bug Fix: When using SSH's X11 forwarding, or due to other internal uncertainties, there may be repeated handshakes and key confirmations between the client and the server.

Related Issue：
-  #8953
-  #8934

image


cfs4819:
image
The building problem doesn't appear to be caused by the latest committed code.

Issue #8957, opened 2023-09-08
Docking does not respect "current" setting with no browser windows open in workspace


Tbhesswebber:
This was happening in v1.0.197, which prompted me to upgrade and see if the issue was fixed.  My system configuration has multiple workspaces on my left (non-primary) display and a single, static workspace on my right (primary) display.  My keybinding is set to `F12` and I have tried some others to see if something is clobbering the keybinding somehow.

Initially, I thought that it was an issue in workspaces with no open applications, but then I opened Software Center and tried the hotkey to no avail.  As far as I can tell, it works on workspaces with a browser window open (electron or otherwise) - Slack, Spotify, VSCode, Chrome, Firefox (weird...)

Please feel free to ask for more specific information and I'd be happy to dig deeper into this weirdness.

Version: 1.0.200
Platform: Linux x64 6.2.0-31-generic
Plugins: none
Frontend: xterm-webgl

Tbhesswebber:
Attempting some additional applications, I discovered that it's actually when an application with a browser has focus, regardless of mouse location.

Some test cases:
- VSCode is focused and mouse moves to empty workspace on other display [WORKS]
- VSCode is focused and mouse moves to empty workspace on other display and clicks [DOESN'T WORK]
- Chrome is focused and open to half screen in workspace and mouse moves over desktop [WORKS]
- Chrome is open to half screen in workspace and mouse clicks on desktop [DOESN'T WORK]

jgauna:
Same issues here

```
Linux j 5.15.0-86-generic #96-Ubuntu SMP Wed Sep 20 08:23:49 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

liushuai05:
> In debugging this, I discovered that `/app/lib/app.ts#139` never gets called when a chromium-based window doesn't have focus. Digging deeper, I think [this Reddit post](https://www.reddit.com/r/gnome/comments/qye01v/why_screen_sharing_and_keyboard_shortcuts_dont/) is actually the culprit. TL;DR: Certain operating systems disallow global key listeners, so users will need to listen to, and forward, the events manually from their OS settings.
> 
> I _kind of_ got around this by setting a custom shortcut to run `tabby`, but it just forces the window open rather than toggling it.
> 
> @Eugeny - is there a way to toggle the window from the command line? Or can this be added in somehow?

[...]

PR #9571, opened 2024-04-02
docs(hacking): adds clarification for installing forks

Tbhesswebber:
In attempting to get things running locally to debug #8957, I ran into an error that was unintuitive to resolve because `yarn` is usually agnostic of git.  I don't have the screenshot, but the solution was to pull the tags from the repo.  This PR adds a note about that.


Eugeny:
Thanks!

PR #9624, opened 2024-05-07
add setting to hide tray (menubar for macos)

### LINKED ISSUE #4955: Feature request: Option to hide the MacOS menu icon
Please add an option to let users to hide it. There are no much room left on my menu bar now :(
image

## Funding this issue
* You can sponsor this specific effort via a [Polar.sh](https://polar.sh) pledge below
* We only receive your pledge once the issue is resolved & you've verified it.

<a href="https://polar.sh/Eugeny/tabby/issues/4955">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/Eugeny/tabby/issues/4955/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

C41M50N:
fixes #4955 (I know it's marked as completed, but it was never implemented from what I can see)

### DIFF
--- a/tabby-settings/src/components/settingsTab.component.pug
+++ b/tabby-settings/src/components/settingsTab.component.pug
@@ -102,15 +102,6 @@
 
                     .form-line(*ngIf='hostApp.platform !== Platform.Web')
                         .header
-                            .title(translate) Show tray
-                            .description(translate) Show Tabby in tray or menu bar.
-                        toggle(
-                            [(ngModel)]='config.store.showTray',
-                            (ngModelChange)='saveConfiguration(true)'
-                        )
-
-                    .form-line(*ngIf='hostApp.platform !== Platform.Web')
-                        .header
                             .title(translate) Debugging
 
                         button.btn.btn-secondary((click)='hostWindow.openDevTools()')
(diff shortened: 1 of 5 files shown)


tabby-settings/src/components/settingsTab.component.pug
```diff
@@ -100,6 +100,15 @@
                             .description(translate) Enable automatic installation of updates when they become available.
                         toggle([(ngModel)]='config.store.enableAutomaticUpdates', (ngModelChange)='saveConfiguration()')
 
+                    .form-line(*ngIf='hostApp.platform !== Platform.Web')
```
Eugeny, line 103: should exclude Linux as well since it's excluded in the backend

also, the setting should be in the Window settings tab

Eugeny:
Sorry for the delay, I'm completely overwhelmed with work right now. Looks good, left 2 notes

C41M50N:
No worries, thanks for taking the time. I made changes to accomodate those notes. 
Another issue that I am seeing is that the 'showTray' config is not added to the config file if it doesn't exist (and is therefore out-of-sync with the toggle). Is there a standard way of handling the case where a config option is added if it is not already set?

Eugeny:
No, unfortunately the main process doesn't have access to the same merged "defaults + user settings" object, so all access to `configStore` has to have a `?? default` slapped onto it.

C41M50N:
Okay. I've made that change as well as shifted to 'hide tray' instead of 'show tray'. That way the default will be false. Let me know if there's anything else that needs changes.

Eugeny:
LGTM, thanks! 

@all-contributors please add @C41M50N for code

Issue #9656, opened 2024-05-18
CLI not available in linux installation


RiccardoManzan:
### Discussed in https://github.com/Eugeny/tabby/discussions/9647

On linux, the `tabby` command does not support the cli usage (like `tabby run <executable>`, `tabby open <folder>` ...)

RiccardoManzan:
PR submitted

PR #9657, opened 2024-05-18
Eugeny#9656 fix: cli usage on linux

RiccardoManzan:
Hi @Eugeny, here it is the fix that should address the bug.
#9656


Eugeny:
Thank you!

Eugeny:
@all-contributors please add @RiccardoManzan for code

PR #9749, opened 2024-07-01
wsl: Get the distro icon from the package family name

microhobby:
All the WSL distributions installed from the Microsoft Store have a package family name that can be used to get the distribution icon.

For example, this way even distributions that are not in the Tabby SVG list can show the distro icon:

image

### DIFF
--- a/tabby-electron/src/shells/wsl.ts
+++ b/tabby-electron/src/shells/wsl.ts
@@ -6,6 +6,8 @@ import { HostAppService, Platform, isWindowsBuild, WIN_BUILD_WSL_EXE_DISTRO_FLAG
 
 import { ShellProvider, Shell } from 'tabby-local'
 
+import { PowerShell } from 'node-powershell'
+
 /* eslint-disable block-scoped-var */
 
 try {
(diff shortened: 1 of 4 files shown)


tabby-electron/src/shells/wsl.ts
```diff
@@ -49,6 +51,14 @@ export class WSLShellProvider extends ShellProvider {
             return []
         }
 
+        // make sure that this will not use the powershell profile
+        // that may take a long time to load
+        const pwsh = new PowerShell({
```
Eugeny, line 56: Move into the constructor to avoid spawning a new pwsh every time the profiles menu is opened

tabby-electron/src/shells/wsl.ts
```diff
@@ -58,7 +68,24 @@ export class WSLShellProvider extends ShellProvider {
...
         if (lxss?.DefaultDistribution) {
             const defaultDistKey = wnr.getRegistryKey(wnr.HK.CU, lxssPath + '\\' + String(lxss.DefaultDistribution.value))
+            let _icon = wslIconMap.Linux
             if (defaultDistKey?.DistributionName) {
+                // check if the register has PackageFamilyName
+                if (defaultDistKey.PackageFamilyName) {
+                    // get the icon from the package family name
+                    const packageFamilyName = (defaultDistKey.PackageFamilyName.value as string).split('_')[0]
+
+                    if (packageFamilyName) {
```
Eugeny, line 78: Extract the common steps into a function here

Eugeny (changes requested):
Just a few comments, otherwise LGTM, thanks :+1:

tabby-electron/src/shells/wsl.ts
microhobby: Done

tabby-electron/src/shells/wsl.ts
microhobby: Done

Eugeny:
Thank you 🚀

Issue #9378, opened 2024-01-08
It's lagging.


Amaderis:
Terrible lag when typing on the server. But in another SSH client (MobaXterm) everything is smooth and without such delays. It is impossible to work on data center nodes at all. I really like this application, better than MobaXterm. But this problem is very stressful. 

Version: 1.0.205
Platform: Windows x64 10.0.22631
Plugins: none
Frontend: xterm-webgl

atlas-eric:
V1.0.205
The large prompt lag on my new installed M2-Mac
One tab open only

Makes it barely useable.

gitmatters:
Use version 1.0.198 . It's the last version, where ssh and x11 forwarding working flawlessly.

element0xE:
> Use version 1.0.197 . It's the last version, where ssh and x11 forwarding working flawlessly.

I can confirm the typing expirience tru SSH is flawless with version 1.0.197
MacOS Sonoma 14.3.1 | M2 Max

atlas-eric:
Correct.
I went back too and it works great. Sonoma both Intel and M2 versions

Eugeny:
Could you please [record a performance trace](https://developer.chrome.com/docs/devtools/performance/reference#record-runtime) on the new version using the built-in debug tools?

atlas-eric:
How do I connect the Tabby app to the Chrome Dev Tools, for that
performance trace?

Eugeny:
Tabby has them built in, it's in Settings -> Application -> Debugging

Eugeny:
That's weird - it works for me via the toolbar button:

<img width="306" alt="Screenshot 2024-03-11 at 21 19 03" src="https://github.com/Eugeny/tabby/assets/161476/2f694961-d441-4406-a5d7-86038267e07f">

Eziou:
Same, v1.0.207 and severe lag on my new m3 pro macbook, macos Sonoma 14.4.

element0xE:
> That's weird - it works for me via the toolbar button:
> 
[...]

Issue #9716, opened 2024-06-17
Poor X11 performance


YaPaY:
I am using 1.0.207 on Windows 11 and I connect to many Linux machines on the same network. My problem with X11 is that the performance is very bad. Even a click can take a few seconds. Putty on the same machine does not have this problem. XMing and VcXsrv both have the same problem.

gitmatters:
This issue is already reported in #8968 and #9378 .

Issue #8968, opened 2023-09-11
X11 Forwarding massivly damaged since version 1.0.198


gitmatters:
Version: 1.0.198 + 
Platform: Windows x64 10.0.19045
Plugins: sftp-tab
Frontend: xterm-webgl

Here is the thing the X11 forwarding in a ssh session under version 1.0.198 is slow and lagging (so slow that practically you cannot use GUI programs anymore, since the simpliest mouse clicks will take seconds to be proceeded). Whats even worse, that the version also affects alternative terminals like Putty (so they than have the exact same problem). Switching tabby back to version 1.0.197 resolves the issue, so it seems like the update made to xterm in 198 has caused the damage.   

The issue is reproducable, a colleague of mine verified (on Windows 11).

gitmatters:
Could anybody reproduce? 
Its still an issue for me with v205...

YaPaY:
weird, still no solution

Eugeny:
Is likely fixed in 209+

PR #9768, opened 2024-07-09
SSH Lag Fix

Vadhvis:
Fixes Eugeny/tabby#9378
Could not test if it helps with Eugeny/tabby#9716 and Eugeny/tabby#8968


Eugeny:
Thank you! And @mscdex thanks for jumping in with the tip

element0xE:
Thanks @Vadhvis !

Issue #8461, opened 2023-05-21
opening the SFTP tab with a hotkey


jayagami:
**Is your feature request related to a problem? Please describe.**
open sftp is inconvenient 

**Describe the solution you'd like**
Please add support for opening the SFTP tab with a hotkey.

**Describe alternatives you've considered**
Please add the ability to pin the SFTP tab.

**Additional context**
Please provide support for the SFTP protocol and enable the option to open it in a new tab.

hemin110:
+1

aqyuki:
I would like to work on this Issue. Can I get some support?

PR #9786, opened 2024-07-13
Fix Issue #8461: Add Toggle Hotkey and Pin Button for SFTP Panel

GeminiLn:
This PR addresses Issue #8461 by introducing two new features for the SFTP Panel:

1. Toggle Hotkey:

- Added a hotkey to toggle the visibility of the SFTP Panel, enhancing user convenience.
Hotkey

2. Pin Button:

- Introduced a Pin button in the SFTP Panel. When activated, this button prevents the panel from closing when clicking outside of it, ensuring that the pinned panel remains open.
Pin

### DIFF
--- a/tabby-core/src/hotkeys.ts
+++ b/tabby-core/src/hotkeys.ts
@@ -255,6 +255,10 @@ export class AppHotkeyProvider extends HotkeyProvider {
             id: 'pane-decrease-horizontal',
             name: this.translate.instant('Decrease horizontal split size'),
         },
+        {
+            id: 'open-sftp-pane',
+            name: this.translate.instant('Open SFTP panel'),
+        }
     ]
 
     constructor (
--- a/tabby-core/src/hotkeys.ts
+++ b/tabby-core/src/hotkeys.ts
@@ -256,8 +256,8 @@ export class AppHotkeyProvider extends HotkeyProvider {
             name: this.translate.instant('Decrease horizontal split size'),
         },
         {
-            id: 'open-sftp-pane',
-            name: this.translate.instant('Open SFTP panel'),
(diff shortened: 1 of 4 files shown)


Eugeny:
Thank you for contributing! In which case is it necessary to pin the panel? 

Having to navigate to the current SFTP folder can be easily avoided by correctly setting up the CWD reporting, and the terminal in the background can't be currently properly used anyway while the panel is open.

GeminiLn:
> Thank you for contributing! In which case is it necessary to pin the panel?
> 
> Having to navigate to the current SFTP folder can be easily avoided by correctly setting up the CWD reporting, and the terminal in the background can't be currently properly used anyway while the panel is open.

Issue #8461 did not specify when it is necessary to pin the panel. In my experience, I sometimes accidentally click outside the panel, causing it to close, which is quite annoying. I think adding a pin feature can prevent this.

And yes, I do notice that the terminal is unusable when the panel is open. I’m curious if it’s possible to adjust the panel to allow terminal use alongside SFTP. If this adjustment is feasible, I’d be willing to work on it sometime.

Eugeny:
TBF I think that would be a more viable improvement, as I don't quite see what this PR is solving (it's trivial to reopen the panel after closing it). 

That would involve converting the `SFTPPanel` component into an `SFTPTab` and changing the SFTP button logic to instead opening a split pane (same as the "split > bottom" menu item) that would host the new SFTP tab sharing the same connection profile.

barrelltitor:
For my workflow it's incredibly annoying to not have a pinned sidebar, would be great to be able to have both the terminal and the sftp pinned somehow, though a hotkey to open it would be great until then

Issue #9389, opened 2024-01-14
Alpha 205 might have broke "vibrancy" -- It's no longer translucent


LBYPatrick:
**Describe the problem**:
That in 204, the gaussian blur glass texture for the window under ``macOS`` works if I turn on "vibrancy". But in 205, this feature has no effect. Is there a breaking change with the introduction of the ``background`` plugin?

**To Reproduce**:
Simply upgrade to 205. 

My machine specs:

Per ``neotech``:

```bash

OS: macOS 14.2.1 23C71 arm64 
Host: MacBookPro18,3 
Kernel: 23.2.0 
Uptime: 3 days, 19 hours, 8 mins 
Shell: zsh 5.9 
Resolution: 1512x982 
Terminal: Tabby 
GPU: Apple M1 Pro 
Memory: 2815MiB / 16384MiB 
```

And of course: 

<img width="297" alt="image" src="https://github.com/Eugeny/tabby/assets/23201539/f6c85254-7408-4c1b-a15b-8f764f91fa11">

jukfiuune:
Same issue on 207, for some reason while turning the feature off the blur appears for a few seconds. Also i tested on Linux, and can say that this is a macOS issue specifically! (AMD64 btw, not arm)

Neofetch:
OS: macOS 13.6.1 22G313 x86_64 
Host: Hackintosh (SMBIOS: iMac18,1) 
Kernel: 22.6.0 
Uptime: 1 day, 9 hours, 3 mins 
Packages: 173 (brew) 
Shell: zsh 5.9 
Resolution: 1920x1080 
DE: Aqua 
WM: Quartz Compositor 
WM Theme: Blue (Dark) 
Terminal: Tabby 
CPU: Intel i3-7100 (4) @ 3.90GHz 
GPU: Intel HD Graphics 630 
Memory: 5851MiB / 8192MiB

jukfiuune:
So what's going on, is this going to be fixed?

fireblue:
Hi, I had the same problem, googled and came across your issue.

Seems like upgrading to macOS 14 causes the problem.

I've made a fix in #9896

LBYPatrick:
tysm! I will try to make a build on my machine based on your PR and will keep you posted on the results!

Eugeny:
@LBYPatrick you can grab the latest nightly build here, the fix should already be in it :) https://nightly.link/Eugeny/tabby/workflows/build/master

pr11me:
[...]

PR #9896, opened 2024-08-22
Fixed the issue where Windows Vibrancy was not working on newer versions of macOS due to the outdated macos-release package.

fireblue:
Hi, 

In current latest version, window vibrancy option in macOS is broken, as mentioned in issue #9389 

After some investigation, I've found the root cause: the version of the macos-release package being used is 3.1.0, which does not support macOS 14. 

As a result, on macOS 14, the `macOSRelease().version` in `app/lib/windows.ts` always returns an empty string, causing the macOSVibrancyType variable to always default to `'dark'` (which has been deprecated in newer versions of macOS) instead of the intended `'under-window'`. 

This ultimately leads to the failure of the `Settings/Window/Vibrancy` option (the background remains totally opaque). 

Updating the macos-release package to the latest version will return the correct system version and resolve this issue.

The attachment shows a comparison before and after the modifications. The NSVisualEffectView used by Electron has relatively high opacity, making the effect less noticeable, but upon close inspection, the difference can be seen.

There seems to be another issue: the `Vibrancy` and `Opaque` options seem to conflict with each other on macOS 14. 

The `Vibrancy` option only takes effect when the `Opaque` option is set to full.

截屏2024-08-22 22 20 45
截屏2024-08-22 22 20 20

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "html-loader": "4.2.0",
     "json-loader": "^0.5.7",
     "lru-cache": "^6.0.0",
-    "macos-release": "^3.1.0",
+    "macos-release": "^3.3.0",
     "ngx-toastr": "^16.0.2",
     "node-abi": "^3.65.0",
     "npmlog": "6.0.2",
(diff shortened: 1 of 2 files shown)


Eugeny:
Thank you!

PR #10015, opened 2024-10-20
Feat#10008: add hotkeys to open profile selectors for a specific group

### LINKED ISSUE #10008: profile-selectors hotkey for groups
Currently these are the hotkeys in Tabby for profile-selector

image

It is restrictive to a particular connection type. While individual connection/saved layout can be assigned a hotkey when opening as new tab, but groups (with a set of connections and/or saved layouts mixed up) does not have a hotkey to open as profile-selector.

For example, I have a group set up for a particular project or work with some different connections and saved layouts. I want to avoid opening `profile-selector` with a large number of options but only for a particular group which I set the needed connections/layouts.

Clem-Fern:
Close #10008: Add hotkeys to open profile selectors for a specific group.

image
image

### DIFF
--- a/tabby-core/src/configDefaults.yaml
+++ b/tabby-core/src/configDefaults.yaml
@@ -31,6 +31,8 @@ hotkeys:
     __nonStructural: true
   profile-selectors:
     __nonStructural: true
+  group-selectors:
+    __nonStructural: true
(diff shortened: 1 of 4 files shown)


Eugeny:
Thank you!

PR #10043, opened 2024-11-05
Fix Issue #10012 - better ssh config parsing

### LINKED ISSUE #10012: ssh config file parsed incorrectly with profiles
Hi,

I have a problem with the last build v1.0.215. The user set in the **.ssh\config** is not used.
I use the ssh config file **C:\Users\xxx.ssh\config** on Windows 11 23H2 22631.4137.

```
Host srv-a
    Hostname 172.22.0.1

Host srv-*
    User admin
    IdentitiesOnly yes
    IdentityFile ~/.ssh/mysshkey
```

When i select the profile **srv-a** imported from **.ssh\config** the user is not **admin** but **root** (i see this on the top of the screen root@172.22.0.1:22).
On a new terminal in Tabby, if i run ssh srv-a the user used is admin.
The problem seems to be located on the parsed elements because when i add **User admin** in the section of srv-a the user used is good.

andylaw:
Code to parse the SSH Config file and convert all usable (non-wildcard name, at least a Hostname specified) configurations into Tabby Profiles. Fixes #10012 and addresses discussion in #10017


Eugeny:
LGTM, thanks!

PR #10076, opened 2024-11-30
Using ssh default profile(user/password) without host to access multiple servers

### LINKED ISSUE #6156: Add centralized authentication method
Please add centralized authentication method for use same credential with some devices, and store it in Vault.

Allow modify password for centralized authentication method from Vault.

image

image

### LINKED ISSUE #10077: Is there a way to have a "password profile" (i.e. central password) for use in SSH sessions ?
### Discussed in https://github.com/Eugeny/tabby/discussions/9669

<sup>Originally posted by **AustinHarisch** May 23, 2024</sup>
I SSH to switches all the time, and I have to configure my password for each and every switch that I connect to. That password is managed by a TACACS+ server, which means if I change my password, I'd have to update it for every switch I have an entry for. Is there a way to have a "password profile" that I can apply to every SSH connection so I'd only have to update it in one spot? i.e. I'd create a new profile with my TACACS+ username and password and call it "TACACS Credentials" or something like that. Or, if that isn't possible, is there at least a way for me to edit the config in bulk so I can change all of the passwords at once?</div>

ianaflous:
**Issue** 
Connecting with SSH to multiple servers that have centralized authentication (same user/password) forces us to save this password in all the profiles (hosts) of our vault, making password changes tedious.
Closes #10077, resolves #6156, resolves #9305, closes #9669

**Solution** 
Now we can use a default profile to create a user/password without specifying the host in the vault. It will be used by default if there's no other entry in the vault with the host we want to connect to.
image

### DIFF
--- a/tabby-core/src/services/vault.service.ts
+++ b/tabby-core/src/services/vault.service.ts
@@ -195,7 +195,13 @@ export class VaultService {
         if (!vault) {
             return null
         }
-        return vault.secrets.find(s => s.type === type && this.keyMatches(key, s)) ?? null
+        let vaultSecret = vault.secrets.find(s => s.type === type && this.keyMatches(key, s))
+        if (!vaultSecret) {
+            // search for secret without host in vault (like a default user/password used in multiple servers)
+            key['host'] = null
+            vaultSecret = vault.secrets.find(s => s.type === type && this.keyMatches(key, s))
+        }
+        return vaultSecret ?? null
     }
 
     async addSecret (secret: VaultSecret): Promise<void> {


Eugeny:
Thank you and sorry for the delay

Eugeny:
@all-contributors add @ianaflous for code

gkostoulias:
I have not been able to successfully get this solution to work on Mac 15.6.1. I am using build 1.0.229.

ianaflous:
This discussion can be helpful #10563.

gkostoulias:
Unfortunately, that was the initial post I saw that forced me down this rabbit hole. No matter what combination of config I seem to do, I can not get it to work. At one point, I did manage to delete all the user/pass I had set individually, still trying to recover from that mess.

PR #10105, opened 2024-12-23
Update sshImporters.ts to parse Include directive

### LINKED ISSUE #10099: .ssh/config folder problem
When you configure your .ssh/config with the following

Include ~/.ssh/config.d/*.conf

and add a folder config.d to organize multiple config files for ssh with multiple hosts

config-develop.conf
config-productive.conf

tabby does not read the config correctly. No host is added to the connections&profiles tab from this path.

Please make default ssh config path configurable or something similar.

Tested and reproduced under Linux and Windows.

Plain terminal works.

xhiroga:
Resolve #10099

I've tested these changes locally and so far they work. However, I'm not entirely sure if this is the best approach, so I'd really appreciate any guidance or suggestions. I'm totally open to revising the code based on your feedback. Thanks in advance!

### Normal cases

- [x] Using wildcards in Include directives
- [x] Using absolute paths in Include directives
- [x] Using relative paths in Include directives
- [x] Handle specifying the same file multiple times in an Include directive
- [x] Handle Includes written within a Match or Host block

### Error handling

- [x] Cyclic references in Include directives can continue to be processed.
- [x] Processing can continue even if the file specified in the Include directive does not have permissions.
- [x] Processes can continue even if the file specified by an Include directive does not exist.
- [x] Processes can continue even if the file specified in an Include directive has an invalid format.

### DIFF
--- a/tabby-electron/src/sshImporters.ts
+++ b/tabby-electron/src/sshImporters.ts
@@ -106,56 +106,43 @@ async function parseSSHConfigFile (
     try {
         raw = await fs.readFile(filePath, 'utf8')
     } catch (err) {
-        console.error('Error reading SSH config file:', err)
-        // Return an empty parsed config on error
+        console.error(`Error reading SSH config file: ${filePath}`, err)
         return SSHConfig.parse('')
     }
 
-    const config = SSHConfig.parse(raw)
-
-    // Make a shallow copy of `config` first to avoid mutating during iteration
-    for (const entry of [...config]) {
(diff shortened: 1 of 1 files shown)


Eugeny:
Tested it and it's looking good, thank you!

xhiroga:
@Eugeny Thanks for verifying it works so far! I’d love to hear your thoughts or suggestions whenever you can. Thank you!

Eugeny:
LGTM 🚀

PR #10138, opened 2024-12-30
Create Hungarian translation

T0liver:
Copied the translation and added Hungarian translation because it was missing.

Please review but I was trying to make as good as it can be and keep the structure of the phrases but it's not always mirror-translation or is the exact same phrase but keeping in mind that the English and Hungarian is not the same this is a pretty good translation.

### DIFF
--- a/locale/hu-HU.po
+++ b/locale/hu-HU.po
@@ -1,0 +1,2574 @@
+msgid ""
+msgstr ""
+"Content-Type: text/plain; charset=UTF-8\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"X-Crowdin-Project: tabby\n"
+"X-Crowdin-Project-ID: 493349\n"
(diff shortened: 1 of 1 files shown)


T0liver:
I've just made a little fix and it works 😁

Eugeny:
Thank you for your contribution but _please_ see https://github.com/Eugeny/tabby/blob/master/locale/STOP.txt

If you can manually merge it with the existing translations, I can upload it to Crowdin: https://gist.github.com/Eugeny/c7f1285d56203c147e4c692a2803485e

T0liver:
Thanks for the warn, I totally missed STOP.txt, sorry.

Merging means now that fill the blank spaces, as I guess right? I'll try and make a commit for this PR :)

Thank you!

Eugeny:
Just give the existing translation a quick look and combine the two so that you keep the best parts of both

PR #10177, opened 2025-01-09
Set the application's dark mode to follow the app settings on macOS.

fireblue:
Hi, 

Last time I fixed the vibrancy issue on macOS: #9899 
Somehow I still feel something strange in the vibrancy effect on macOS.

After some digging, I've found that the translucent effect of an Electron app's background on macOS is implemented by using NSVisualEffectView, which adapts to the system's dark mode setting in default—black in dark mode and white in light mode. 

Tabby did not set the app-level dark mode configuration, so it defaults to following the system's dark mode setting. 

As a result, when the macOS system dark mode is set to light, Tabby's dark theme background will mix with a white background, and appears washed out. 

I updated the code to make Tabby's dark mode follow the app's internal settings, and dark themes look better.

I made two screenshots to show the difference with the following custom css:

```css
terminal-toolbar {
    opacity: 0.8 !important;
    background: rgba(255, 255, 255, 0.1) !important;
    backdrop-filter: blur(10px) !important;
    -webkit-backdrop-filter: blur(10px) !important;
}
.tabby {
    opacity: 0.75 !important;
}
```

The old code has the same effect with "always light"
截屏2025-01-09 18 26 26

The new code has a darker and more translucent(which appears) background:
截屏2025-01-09 17 12 06

I think we can ultimately achieve a near-transparent black glass like effect by configuring the material and opacity of the NSVisualEffectView, but Electron does not expose an interface for us to modify these parameters. Will try to dig further.

### DIFF
--- a/tabby-ssh/src/session/ssh.ts
+++ b/tabby-ssh/src/session/ssh.ts
@@ -37,7 +37,6 @@ type AuthMethod = {
     type: 'publickey'
     name: string
     contents: Buffer
-    hashAlg: 'sha256'|'sha512'|'sha1'|null
 } | {
     type: 'agent',
     kind: 'unix-socket',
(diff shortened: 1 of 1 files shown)


fireblue:
Currently the dark mode is applied to macOS only because I'm worried about the possiblity of breaking something else on other platforms.

Eugeny:
Tested it and it looks good :+1:
However you can't import `electron` from plugins other than `tabby-electron` as it will break tabby-web.

You could send the message from `updateVibrancy()` in tabby-electron instead

fireblue:
> Tested it and it looks good 👍 However you can't import `electron` from plugins other than `tabby-electron` as it will break tabby-web.
> 
> You could send the message from `updateVibrancy()` in tabby-electron instead

roger that, fixing.

PR #10186, opened 2025-01-10
Set the application's dark mode to follow the app settings on macOS.

fireblue:
Hi, 

This pull request is a continuation of #10177 

Your **elegant** solution completely eliminates the need to modify colorSchemeSettingsTab.component.ts/pug👍 👍 

You’re amazing!

To avoid unnecessary back-and-forth changes to these two files, I’ve followed your instructions and made the adjustments. This new pull request reflects those updates.

Thanks again for your guidance!

### DIFF
--- a/app/lib/window.ts
+++ b/app/lib/window.ts
@@ -1,7 +1,7 @@
 import * as glasstron from 'glasstron'
 import { autoUpdater } from 'electron-updater'
 import { Subject, Observable, debounceTime } from 'rxjs'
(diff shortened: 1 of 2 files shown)


fireblue:
Cloned Electron source code and took a look at native_window_mac.mm, the code of windows vibrancy is quite straight forward, everything is initialized and configured in function 
```cpp
void NativeWindowMac::SetVibrancy(const std::string& type)
```
but there's no much to customize with NSVisualEffectView.
To achieve a more translucent and customizable background, implementing yet another glass like blurring background view should do the trick, but that's too heavy and possibly problematic in every way. :(

For now, current approach might be the best solution.

(From a guy who is obsessed in translucent glass backgrounds)

Eugeny:
Great, thanks!

PR #10196, opened 2025-01-13
Keep the translucency effect even when the window loses focus on macOS.

fireblue:
Hi,

Changed a small configuration on macOS, so the window will remain translucent even it loses focus.

I think it might look better?

截屏2025-01-13 18 16 03

截屏2025-01-13 18 16 26

Sorry, I’m not very familiar with Electron, so I end up modifying just a tiny little part of the code each time to fix an issue I find while using Tabby. lol

Thank you!

### DIFF
--- a/app/lib/window.ts
+++ b/app/lib/window.ts
@@ -101,6 +101,10 @@ export class Window {
         }
 
         if (process.platform === 'darwin') {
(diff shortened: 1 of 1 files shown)


Eugeny:
Looks great, thanks 👍

PR #10308, opened 2025-02-15
Fix: Unable to launch WinSCP for SSH sessions using private key

### LINKED ISSUE #10113: [v1.0.216] Cannot launch WinSCP
**Describe the problem**:
Cannot launch WinSCP on lastest release v1.0.216

**To Reproduce**:
Use shortcut or right-click menu

OpaqueGlass:
This PR attempts to fix #10113.

Read the location of the privateKey from `SSHSession.option.privateKeys` and pass the passphrase to `winscp.com /keygen`.

Tested scenarios:

* the privateKey from the file system or the vault.
* Read the saved password from passwordStorage.

Unsupported scenario:

* If the privateKey is set with a passphrase and the passphrase is not remembered, it will not prompt the user for the password and will throw an error at the /keygen step.

  To accomplish this, it may be necessary to modify or duplicate some code in `session/ssh.ts` `SSHSession.loadPrivateKeyWithPassphraseMaybe()`. Considering my implementation is a bit repetitive with some logic in `ssh.ts` , I haven't made modifications about this scenario.

### DIFF
--- a/tabby-ssh/src/services/ssh.service.ts
+++ b/tabby-ssh/src/services/ssh.service.ts
@@ -1,7 +1,8 @@
-// import * as fs from 'fs/promises'
+import * as fs from 'fs/promises'
+import * as crypto from 'crypto'
 import * as tmp from 'tmp-promise'
 import { Injectable } from '@angular/core'
-import { ConfigService, HostAppService, Platform, PlatformService } from 'tabby-core'
+import { ConfigService, FileProvidersService, HostAppService, Platform, PlatformService } from 'tabby-core'
 import { SSHSession } from '../session/ssh'
 import { SSHProfile } from '../api'
 import { PasswordStorageService } from './passwordStorage.service'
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you!

Eugeny:
@all-contributors add @OpaqueGlass for code

patildhruv:
@Eugeny @OpaqueGlass I tried adding that code to my Installation files, but still not working.

Do I need to create a dev env, install dependencies using npm, ng serve, and check?
Please help me out as this feature is mosttttt needed for me.
Any advice in the correct direction would be of utmost help.

OpaqueGlass:
> Do I need to create a dev env, install dependencies using npm, ng serve, and check?

If you want to build it yourself, then yes.  

Alternatively, you can simply roll back to the version before the error occurred (v1.0.215) and wait for the next production release.  

> Another option is to find a recently successful action run and download the packaged content, e.g.:  
[https://github.com/Eugeny/tabby/actions/runs/13584935249](https://github.com/Eugeny/tabby/actions/runs/13584935249)  

**Note:** Using a development build may come with potential risks.

PR #10314, opened 2025-02-20
Packagecloud upload for current Debian releases (bullseye, bookworm, trixie)

madebyTimo:
Currently, Debian deb packages are available only for releases up to buster (10). No package is available for newer releases.

This pull request adds uploads for bullseye, bookworm, and trixie (11–13), which correspond to oldstable, stable, and testing, respectively.


Eugeny:
@all-contributors add @madebyTimo for code

PR #10353, opened 2025-03-07
scroll_terminal_by_line

gh-log:
this feature is preventing me from using Tabby for work.

set lines to scroll to 1 but can be a variable modified in settings.


Eugeny:
@all-contributors add @gh-log for code

PR #10370, opened 2025-03-13
Packagecloud_upload_for_Debian_testing_unstable

gh-log:
I use Debian unstable so would like to update Tabby with apt.


Eugeny:
Thanks!

PR #10468, opened 2025-05-01
fix(ssh/settings): formats date value according to locale rules

Clem-Fern:
Hey Eugene, 

Asked in Clem-Fern/rtabby-web-api#29, Formats date value according to locale rules.
I've only found 3 places in which date are displayed (release notes, FTP, config sync).

EN
image

FR now
image

FR before
image


Eugeny:
Could you please wrap it in a new custom date pipe in `tabby-core` instead? This will make it easier to use later.

Clem-Fern:
> Could you please wrap it in a new custom date pipe in `tabby-core` instead? This will make it easier to use later.

I'm not sure about the pipe name, does it look ok to you ?

Eugeny:
Yup it's ok, thanks again :)

PR #10501, opened 2025-05-21
chore: various major dependency updates

### LINKED ISSUE #10500: Electron 32 is EOL
**Describe the problem**:
This project is still on Electron 32, which is bad for security reasons, as it's been EOL for 2.5 months now. I reckon a simple `package.json` bump will do.

Because EOL electron versions aren't shipped there, we are currently having difficulties packaging this project for [NixOS](https://nixos.org), re: https://github.com/NixOS/nixpkgs/pull/408724 and https://github.com/NixOS/nixpkgs/pull/368048

**To Reproduce**:
- https://endoflife.date/electron
- Click "show more unmaintained releases"
- EOL 04 Mar 2025

dtomvan:
Fixes #10500

Maybe monkeypatching in the upstream `unix/pty.cc` wasn't a good idea; the output app hangs after loading screen... No idea why, don't know how to debug that.

But I think this is a good foundation to start with nontheless... just trying to help out here :_


dtomvan:
I just moved the actions to corepack... at least this way `act` can get to actually running yarn, which it couldn't before anymore because yarn v1 is of course incompatible with yarn v4.

Eugeny:
Sorry but this PR makes no sense to me as it is right now. You've downgraded node-pty to 0.12 (?!), replacing it with a binary build but then throw in a bunch of source patches for it (?) that won't be used because it's already prebuilt.

dtomvan:
This PR has fallen out of date. I'll close.

Issue #10034, opened 2024-10-29
Suggest entering saved password upon sudo.


alessiodam:
**Is your feature request related to a problem? Please describe.**
No, the feature is just a feature :P

**Describe the solution you'd like**
When a connection (for example SSH into another machine) requests for a sudo password, suggest filling in the saved password (if there is one)

**Describe alternatives you've considered**
N/A

**Additional context**
I moved from Termius to Tabby. Tabby is clearly better but I'm missing that one feature that honestly is a must-have. Honestly, Tabby is much better than any terminal program I've used so far, love your work guys!
For example add a dropdown with the suggestion
![image](https://github.com/user-attachments/assets/259df82a-b1fb-430f-9b84-c88fd4d66f44)

witalit:
I am also looking to move to Tabby from Termius.. would love this feature to be added. I got used to Termius with its password suggestion for sudo.

arvkonstantin:
+1

laurogianluca:
+1

gianlucalauro:
+1

Taoshan98:
+1 @Eugeny pls read this.

alessiodam:
Thanks a lot :)

pc-support-services:
Works perfectly on remote Linux but would be nice to have it enabled for remote macOS as well.

PR #10510, opened 2025-05-27
SSH password paste

ralcaidev:
This feature allows users to paste passwords in the terminal of active SSH session, useful when host asks frequently for password.
Ctrl+K is the default keystroke.


Eugeny:
IMO this is more suited to be a plugin rather than a core function - let me know if you need guidance

ralcaidev:
Is there any way someone you know could implement this functionality in a plugin?

Eugeny:
For sudo, it's now implemented in the nightly: #10034

Issue #10681, opened 2025-08-25
Defaults for SSH unset username , can't connect ssh.


hamasa741:
In the advanced settings of the configuration file, I removed the default username for SSH. 
When I connect to a new SSH session, I get the error “Account is required”.
I thought that with ask every time, it would let me enter the username manually each time I connect.
But it's not work.

<img width="726" height="401" alt="Image" src="https://github.com/user-attachments/assets/bab05afe-0348-44c5-882f-4c264669ee52" />

<img width="790" height="565" alt="Image" src="https://github.com/user-attachments/assets/72a417da-1b18-4946-af5a-1d6973a2be86" />

Version: 1.0.226
Platform: macOS arm64 24.6.0
Plugins: title-control, sftp-tab, save-output, highlight, colors-noctis
Frontend: xterm-webgl

RSM-MBI:
Same problem here. I'm connecting to an IBM Hardware Management Console and I'm connecting with different users depending on  what I'm about to do. It would be nice to get a prompt asking for the user so I don't have to create a whole profile for each one. Trying different authentication methods seems to have no effect on this.

Version: 1.0.227
Platform: Windows 10 22H2
Plugins: default
Frontend: default

<img width="657" height="294" alt="Image" src="https://github.com/user-attachments/assets/3cf049e3-5387-4455-b92c-29ae3cf40757" />

<img width="806" height="518" alt="Image" src="https://github.com/user-attachments/assets/e0eb41e5-d4ab-49e4-b4eb-c48d2d578d7f" />


On a sidenote, after trying out multiple terminal clients while looking for something that isn't PuttY, this is the coolest looking and easiest to use by far and I love it. Thank you for doing this.

balzerc:
I am running into the exact same issue, hoping for some guidance here:

<img width="462" height="254" alt="Image" src="https://github.com/user-attachments/assets/05468285-423b-4c28-9735-82ee1d767499" />

onethree7:
I can confirm the same issue,
Tested on Corp and Home machines,
Tested with placeholder and without, the "account is required" is breaking to input a user name.

I need this in my workflow urgently. Why?
We have proxys like example 127.1.1.1 used to connect to a multitude of Customer machines. However, each time we have a different machine or user behind with different credentials in worst case.

That requires config edit before each connection basically making this tool not suitable for quick work, despite its an excellent suite.

onethree7:
https://github.com/Eugeny/tabby/issues/10594
[...]

PR #10786, opened 2025-10-07
Pr askeverytime fix https://github.com/Eugeny/tabby/issues/10681

onethree7:
Summary ->> THIS IS FIXING https://github.com/Eugeny/tabby/issues/10681

The SSH session workflow waits for the prompted username, rehydrates any stored password for that account into the authentication methods, and saves or purges credentials under the correct login when authentication succeeds or fails.

Issue Fix

With the session now keyed to the username collected from the “Always Ask” prompt, users are actually asked for the login when it’s blank, and the subsequent authentication steps reuse (or update) the matching stored password rather than one tied to a different account.

After: prompt appears again:
image

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -30,7 +30,7 @@
     "native-process-working-directory": "^1.0.2",
     "npm": "6",
     "rxjs": "^7.5.7",
-    "russh": "0.1.26",
+    "russh": "0.1.27",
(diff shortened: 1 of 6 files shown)


Eugeny:
Thank you!

Eugeny:
@all-contributors add @onethree7 for code

PR #10803, opened 2025-10-14
fix: adjust ssh config time resolution

osm1892:
OpenSSH's ServerAliveInterval and ConnectTimeout are in seconds, while Tabby handles them as milliseconds.
To fix this incompatibility, the import process now multiplies these values by 1000, converting them from seconds to ms.

### DIFF
--- a/tabby-electron/src/sshImporters.ts
+++ b/tabby-electron/src/sshImporters.ts
@@ -219,7 +219,12 @@ function convertHostToSSHProfile (host: string, settings: Record<string, string 
             case SSHProfilePropertyNames.ReadyTimeout:
                 const secondsString = settings[key]
                 if (typeof secondsString === 'string') {
-                    options[targetName] = parseInt(secondsString, 10) * 1000
+                    const parsedSeconds = parseInt(secondsString, 10)
+                    if (!isNaN(parsedSeconds) && parsedSeconds >= 0) {
+                        options[targetName] = parsedSeconds * 1000
+                    } else {
+                        console.log(`Invalid value for ${key}: "${secondsString}"`)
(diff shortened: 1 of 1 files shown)


Eugeny:
I don't see any runtime changes that your PR could introduce?

osm1892:
> I don't see any runtime changes that your PR could introduce?

The ServerAliveInterval and ConnectTimeout directives in the standard SSH config are originally specified in seconds, and OpenSSH clients interpret them as such. However, Tabby internally handles its corresponding fields—KeepAliveInterval and ReadyTimeout—in milliseconds (ms).

During the config import process, Tabby currently assigns the ServerAliveInterval and ConnectTimeout values directly to its internal fields. This mismatch in units causes problems:

If a user sets a standard parameter like 60 (1 minute) in their OpenSSH config, Tabby treats this as 60ms. Furthermore, if there is a subsequent division by 1000 internally, the field value becomes 0 because the value is less than 1000. This results in a connection error when attempting to connect using the imported SSH profile.

To set a functional value, a user must currently specify a value of 1000 or greater. The core issue is that setting the ServerAliveInterval to, say, 1000 seconds (to satisfy Tabby's ms requirement) virtually disables the heartbeat in any other standard SSH program, as the default SSH server timeout is usually around 300 seconds.

Therefore, to eliminate this unit discrepancy and ensure compatibility, I've added logic to multiply the value by 1000 for these specific fields during the config import process.

Eugeny:
Oh nevermind - I was accidentally looking at a diff of a single commit only - thanks!

Eugeny:
@all-contributors add @osm1892 for code

PR #10931, opened 2025-12-17
export getKeyName for plugin to use

zeroleo12345:
export getKeyName for plugin to use the same approach to get key name

### DIFF
--- a/tabby-terminal/src/frontends/xtermFrontend.ts
+++ b/tabby-terminal/src/frontends/xtermFrontend.ts
@@ -168,7 +168,7 @@ export class XTermFrontend extends Frontend {
             this.hotkeysService.pushKeyEvent(name, event)
 
             let ret = true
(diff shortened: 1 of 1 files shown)


Eugeny:
Changing the `matchActiveHotkey` argument like this will cause all keystrokes from multi-stroke shortcuts to be sent to the terminal, which is bad. Option-o sending ø is normal macOS behaviour when the event reaches the terminal. I've tried reproducing it and if a shortcut is mapped to Option-o, it is correctly not getting passed down to the terminal already.

zeroleo12345:
Thanks @Eugeny to reply, I revert changing `matchActiveHotkey`, and pls review if it okay to export getKeyName.

zeroleo12345:
And actually `matchActiveHotkey` can be reproduce like this, it can only happen once at the key pressed:
1. set hotkey `Option-o` as pasting
2. close Tabby and reopen
3. press `Option-o`

PR #10943, opened 2025-12-25
fix: include http proxy settings in ssh multiplexer key

TheBlindM:
This prevents incorrect session reuse when different HTTP proxies are configured for the same host.


Eugeny:
Thank you! @all-contributors please add @TheBlindM for code

PR #10960, opened 2026-01-10
Fix hiding the app from macos dock when docking is enabled

### LINKED ISSUE #9037: Hide the mac DOCK icon
<!--
My macOS dock is jam-packed, and since I can easily show and hide tabby with option+space, I don't need the dock icon. I forgot exactly where I saw the shell changing functions of the dock icon right click context menu, but it's somewhere else too. So far, I'm not much of a shell switcher anyways.

This issue isn't too unrelated to https://github.com/Eugeny/tabby/issues/4955, but even if that was left unsolved, I have mac Bartender 5 to deal with the menu bar icon more easily. For the dock icon, I don't know of an analogous solution. I apologize if there's a pre-existing method, I didn't find a similar issue on here. 

Just some toggle/checkbox in the tabby settings menu to hide the dock icon. Disabled by default. 

I've tried looking through my frequently used apps and finding a dock hiding option in them, but the dock clutter is still there.

**Additional context**
N/A

SergeBakharev:
Resolves #9037 which is actually a bug. In MacOS it looks like the intent was to hide the application from the MacOS dock when the application is "docked" with the global hotkey, however this didn't work correctly.

Also small improvement to the description of the hideTray config setting to make it clearer that it hides the tray icon.

### DIFF
--- a/app/lib/window.ts
+++ b/app/lib/window.ts
@@ -254,8 +254,12 @@ export class Window {
         if (process.platform === 'darwin') {
             // Lose focus
             Menu.sendActionToFirstResponder('hide:')
+            // Don't disable docked window styles when hiding - keep dock hidden if feature is enabled
             if (this.isDockedOnTop()) {
-                await this.enableDockedWindowStyles(false)
+                // Temporarily disable always-on-top and other properties while hidden
+                if (this.window.isAlwaysOnTop()) {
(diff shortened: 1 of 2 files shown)


Eugeny:
Thanks! @all-contributors please add @SergeBakharev for code

Issue #10959, opened 2026-01-10
Failed to upload the large file.


wangda007:
v1.0.229 - Windows x64 platform: Failed to upload large files, followed by an automatic disconnection of the connection. Even a 68MB file package cannot be uploaded. Please fix this issue promptly!!!

nopeno:
same happens on verion 1.0.229, windows 11

404ZenMaster:
<img width="240" height="308" alt="Image" src="https://github.com/user-attachments/assets/3a1a6072-c02a-43e3-816f-b8a7be7476ae" />
Windows 11 25H2    This version still cannot upload files, uploading files will session closed.
Please fix this issue as soon as possible

ajkrj:
@404ZenMaster @wangda007 @nopeno 

Can you try v1.0.215 and let me know if the file upload works for you then? 

What is your exact use case and how are you uploading the file?

Are you using some kind of tunnel/ports?

nopeno:
i tried again:

   version 1.0.229: failed.
   version 1.0.230: it is ok.

both were using socks5 proxy


> [@404ZenMaster](https://github.com/404ZenMaster) [@wangda007](https://github.com/wangda007) [@nopeno](https://github.com/nopeno)
> 
> Can you try v1.0.215 and let me know if the file upload works for you then?
> 
> What is your exact use case and how are you uploading the file?
> 
> Are you using some kind of tunnel/ports?

999man:
@ajkrj Do I need to completely uninstall and reinstall? I am still unable to upload large files using overlay installation. Version: 1.0.230
Connect to my server using port 22

<img width="582" height="99" alt="Image" src="https://github.com/user-attachments/assets/f5f50a33-3089-4b9a-8ba2-071c19b23702" />

Issue #10224, opened 2025-01-16
X11 forwarding not working since release 1.0.216


mt50su:
**Describe the problem**:
Starting with release 1.0.216, X11 forwarding stopped working

**To Reproduce**:
1, Launch VcXsrv on the local machine.
2, Enable the X11 forwarding option in an SSH profile.
3, Use this profile to establish an SSH session and enable "X11Forwarding" in the sshd_config file on the remote machine.
4, Restart the remote sshd service and reconnect the SSH session.
5, Run an X application, such as xeyes, within the SSH session.
6, With Tabby release 1.0.215, an xeyes window appears as expected. However, with releases 1.0.216 and 1.0.219, nothing happens.

Eugeny:
Should be fixed in the nightly once [this build](https://github.com/Eugeny/tabby/actions/runs/12817954508) passes.

mt50su:
Thanks for your great work, Eugeny.
I am sorry that I forgot to provide the OS information running on the remote machine when I encountered the X11 forwarding issue.
Actually it is "Debian GNU/Linux bookworm 12.9 aarch64" running on Raspberry Pi 5. I tried the latest tabby release 1.0.220, X11 forwarding still did not work in that scenario while 1.0.215 worked perfect.
It will be highly appreciated if this issue can be fixed in the coming tabby releases.

Thanks

Eugeny:
Could you please check the debug logs as in https://github.com/Eugeny/tabby/issues/10237#issuecomment-2608237033 ? Thanks

Issue #10711, opened 2025-09-02
v1.0.227: LOCAL port forwarding broken when accesing a third device


anewbug:
**Describe the problem**:
When using the port forwarding setup in Tabby, the LOCAL forwarding only works 
for listening ports on the server that is being logged into.

It does not work for listening ports on a third server or device.  
This is standard for other ssh clients, an example of openssh is provided below.

the equivalent of the following from a normal openssh client does not work:

**To Reproduce**:

THE FOLLOWING WORKS  FOR TABBY as well as for OPENSSH : :   
Tabby port setup:                                      Forwarded (local) 127.0.0.1:8021 → (remote) 127.0.0.1:80
Equivalent commandline for openssh:      ssh -p 22 -L 8021:127.0.0.1:80 username@server.com  -v

THIS DOES NOT WORK FOR TABBY, BUT WORKS for OPENSSH :  
Tabby port setup:                                      Forwarded (local) 127.0.0.1:8021 → (remote) 192.168.1.21:80
Equivalent commandline for openssh:      ssh -p 22 -L 8021:192.168.1.21:80 username@server.com  -v

Test by opening a browser to:   http://127.0.0.1  on the local machine from where the connection was established.
Assumption, that server at 192.168.1.21 port 80 is a webserver.

anewbug:
Tested on version 1.0.227

Issue #10331, opened 2025-02-27
Tabby is slower, SSH tunnels are lagging, SSH agent connection is broken


arttit:
After the latest update, Tabby has become much, much slower when handling SSH connections.

SSH tunnels are extremely slow, making it difficult to work efficiently.
I tested with another SSH client (Putty), and the tunnels work normally.

SSH agent authentication no longer works, even though it was working fine in the previous version.
` Failed to authenticate using agent: Error: Key(IO(Custom { kind: UnexpectedEof, error: "early eof" }))`

These issues make it hard to use Tabby for SSH connections. 

Version: 1.0.221
Platform: Windows x64 10.0.19045
Plugins: search-in-browser, save-output
Frontend: xterm-webgl

johnalanyoung:
I've been suffering from the same issue for some time with my SSH tunnels being very slow and to the point of not working at all. Putty works without issues to same ssh host.

Tabby was working fine until a certain update.

Version: 1.0.227
Platform: Windows x64 10.0.26100
Plugins: save-output

johnalanyoung:
I've done some testing, it appears this issue creeps in on version: **1.0.221**

1.0.209 - no performance issue
1.0.216 - no performance issue
1.0.220 - no performance issue
1.0.221 - _severe performance issue with SOCKS proxy_

TzunamiDev:
I used to be able to use SSH tunnels perfectly with Tabby. Then I stopped using it for a while and reformatted my computer, so I don't know which version worked well for me.

The point is that now when I try to use SSH tunnels with Tabby, they don't work well. The connection is super slow and unstable. Sometimes it works and sometimes it doesn't. I tested it with PuTTY, and it worked great, but obviously I'd like it to work as well as it did before in Tabby.

johnalanyoung:
Hey, it looks like the above fix has fixed the issue for me! 🥳

PR #10964, opened 2026-01-12
fix(ssh): add EOF handling and error propagation for port forwarding

### LINKED ISSUE #10116: Port forwarding is not working
Hi,

I could not use port forwarding functionality. In the screenshot, I set up a connection forwarding for an RDS instance with 1433 port. RDS instance is only accessible using the ssh server (bastion host).

tabby_forward

At this phase, I could see the port forwarding message in tabby's ssh log:

port_forward

In the same scenario, I was able to connect same instance with bare ssh command:

#ssh -L 1433:some_rds_addr:1433 username@bastion_host_addr

I want to ask is port forwarding feature of tabby has the same functionality of ssh's port forwarding ?

Best Regards

### LINKED ISSUE #10142: MySQL cannot be connected through a tunnel
Version: 1.0.216
Platform: macOS arm64 24.1.0
Plugins: sync-config, sftp-tab, quick-cmds
Frontend: xterm-webgl

After upgrading to the latest version, MySQL cannot be connected through a tunnel, and the error message is as follows:
1159 - Got timeout reading communication packets

It was normal before the upgrade

### LINKED ISSUE #10340: Port forwarding freeze after server resume
**Describe the problem**:
When applications are served on a remote host port, and port forwarding is set, the first load happens normally. If the server is interrupted, then restarted, several things happen:
1. the connection is not severed in the local browser
2. hitting refresh or re-querying `localhost:port` causes the page to hang, endlessly stuck on "loading"

This is specific to version after the switch to `russh`. The only solution I have found is downgrading to pre-russh versions.

**To Reproduce**:
* Install a pre-russh update of tabby
* serve something remotely on port 1234, like a shiny application: `R -e 'shiny::shinyApp(ui=shiny::fluidPage(1234), server=function(input, output, session){}, options=list(port=1234, launch.browser=F))'`
* Connect to the forwarded port, the app will show 1234.
* Ctrl+C the app, then launch another one on the same port: `R -e 'shiny::shinyApp(ui=shiny::fluidPage(4567), server=function(input, output, session){}, options=list(port=1234, launch.browser=F))'`
* This will, as expected, disconnect the first app and load the second one, upon reload it will show 4567.

* Now repeat the same steps with post-russh updates. You'll see the first app not disconnecting and the second reload being stuck in limbo

ajkrj:
Fixes port forwarding issues introduced in russh migration where bi-directional communication would hang

The russh migration (v1.0.216) replaced Node.js Streams with RxJS Observables, requiring explicit EOF (End-Of-File) and error handling. The previous implementation only handled full connection closure but not half-closure (EOF) scenarios, causing applications requiring bi-directional communication to deadlock.

  **Root Cause:**
  In ssh2, EOF was automatically propagated via Stream 'end' events. With russh's Observable-based API, EOF must be explicitly handled via channel.eof$ subscriptions. Without this, when a remote server sent EOF to signal "done sending, waiting for response", the local socket never received notification, causing both sides to wait indefinitely.

  **Changes Made:**

  1. **EOF Propagation (Remote → Local)**
     - Added channel.eof$.subscribe() handlers in both local and remote forwarding
     - Calls socket.end() on EOF to properly half-close the connection
     - Critical for HTTP POST/PUT operations with "100 Continue" responses

  2. **EOF Signaling (Local → Remote)**
     - Added socket.on('end') handlers to call channel.eof()
     - Enables proper half-duplex communication

  3. **Error Handling**
     - Added error callbacks to Observable subscriptions
     - Prevents silent failures and orphaned connections
     - Added comprehensive logging for debugging

  4. **Buffer Conversion Fix**
     - Changed Uint8Array.from(data) to new Uint8Array(data.buffer, data.byteOffset, data.byteLength)
     - The old method created incorrect array structure [98, 117, 102...] instead of proper typed array
     - Provides zero-copy view of buffer data for better performance

  5. **Additional Event Handlers**
     - Added socket.on('error') handlers to properly close channels on socket errors
     - Added debug logging for connection lifecycle events

  **Affected Code:**
  - Lines ~830-880: Local/Dynamic port forwarding (ssh -L and -D)
  - Lines ~500-550: Remote port forwarding (ssh -R)

  **Issues Fixed:**
  - MySQL/MariaDB connections through tunnels (timeout reading packets)
  - Web applications with bi-directional communication
  - Port forwarding freezes after server restart
  - General port forwarding reliability issues

  Fixes #10340
  Fixes #10142
  Fixes #10116
  Related to #10711, #10331

### DIFF
--- a/tabby-ssh/src/session/ssh.ts
+++ b/tabby-ssh/src/session/ssh.ts
@@ -496,48 +496,7 @@ export class SSHSession {
                 channel.close()
             })
 
-            // Channel → Socket data flow with error handling
-            channel.data$.subscribe({
-                next: data => socket.write(data),
-                error: err => {
-                    this.logger.error(`Remote forward channel error: ${err}`)
-                    socket.destroy()
-                },
-            })
-
-            // Socket → Channel data flow with proper conversion
-            socket.on('data', data => {
-                try {
-                    channel.write(new Uint8Array(data.buffer, data.byteOffset, data.byteLength))
-                } catch (err) {
-                    this.logger.error(`Remote forward write error: ${err}`)
-                    socket.destroy()
-                }
-            })
-
-            // Handle EOF from remote
-            channel.eof$.subscribe(() => {
-                this.logger.debug('Remote forward channel EOF')
-                socket.end()
-            })
-
-            // Handle channel close
-            channel.closed$.subscribe(() => {
-                this.logger.debug('Remote forward channel closed')
-                socket.destroy()
-            })
-
-            // Handle socket close
-            socket.on('close', () => {
-                this.logger.debug('Remote forward socket closed')
-                channel.close()
-            })
-
-            // Handle EOF from local
-            socket.on('end', () => {
-                this.logger.debug('Remote forward socket end')
-                channel.eof()
-            })
+            this.setupSocketChannelEvents(channel, socket, 'Remote forward')
 
             socket.on('connect', () => {
                 this.logger.info('Connection forwarded')
@@ -559,19 +518,7 @@ export class SSHSession {
             try {
                 const x11Stream = await socket.connect(displaySpec)
                 this.logger.info('Connection forwarded')
-
-                channel.data$.subscribe(data => {
-                    x11Stream.write(data)
-                })
-                x11Stream.on('data', data => {
-                    channel.write(Uint8Array.from(data))
-                })
-                channel.closed$.subscribe(() => {
-                    socket.destroy()
-                })
-                x11Stream.on('close', () => {
-                    channel.close()
-                })
+                this.setupSocketChannelEvents(channel, x11Stream, 'X11 forward')
             } catch (e) {
                 // eslint-disable-next-line @typescript-eslint/no-base-to-string
                 this.emitServiceMessage(colors.bgRed.black(' X ') + ` Could not connect to the X server: ${e}`)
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you! @all-contributors please add @ajkrj for code

ajkrj:
@Eugeny 
Thank you for fixing the X11 and Agent Forwarding as well. I missed those. 

Few other issues could also be marked as completed based on this fix.  
 Fixes #10224, #10237, #10580 (X11 forwarding not working)
Could be related to #10558, #10131

ajkrj:
#10967 #10959 #10941 #10929 #10928 #10921

ajkrj:
@Eugeny 

I shared the issue with Claude Code and it has suggested the following. I'm not sure how accurate it is but since you are maintaining russh as well I thought it would be good to share it. 

The ssh2 library's stream.pipe() automatically handled backpressure by:
  - Pausing the readable stream when the writable buffer was full
  - Resuming when the writable emitted 'drain'

  The russh migration to RxJS Observables has no backpressure mechanism, causing:
  - Unbounded memory growth during large file transfers
  - Connection crashes/disconnections when buffers overflow
  - Issue #10959 

  Current State:

  The Rust implementation has the necessary APIs:
  - russh/russh/src/client/session.rs:264 - pub fn sender_window_size(&self, channel: ChannelId) -> usize
  - Window adjust events are handled internally

  But the TypeScript bindings don't expose these APIs:
  - app/node_modules/russh/lib/native.d.ts - SshClient class has no window size methods
  - app/node_modules/russh/lib/channel.d.ts - Channel class has no window size observables

  Requested API Additions:

  Option 1: Expose Window Size (Preferred)

  Add to SshClient in the NAPI bindings:
 ```
 export class SshClient {
    // ... existing methods ...

    /** Get the current sender window size for a channel */
    senderWindowSize(channelId: number): Promise<number>
  }

  Add Observable to Channel wrapper:
  export declare class Channel extends Destructible {
    readonly data$: Observable<Uint8Array>;
    readonly eof$: Observable<void>;
    readonly closed$: Observable<void>;
    readonly windowAdjust$: Observable<number>;  // ← NEW: Emits new window size

    // ... existing methods ...
  }

```
  Option 2: Expose Write Backpressure

  Make channel.write() indicate backpressure:
  ```
export declare class Channel extends Destructible {
    // Current:
    write(data: Uint8Array): Promise<void>;

    // Alternative: Return boolean indicating if more data can be written
    write(data: Uint8Array): Promise<boolean>;  // false = pause sending
  }

```
  Implementation Path:

  With these APIs exposed, we can implement proper backpressure:

 ```
 // Socket → Channel with proper backpressure
  socket.on('data', async data => {
      const windowSize = await this.ssh.senderWindowSize(channel.id)

      if (windowSize < 8192) {  // Low window threshold
          socket.pause()

          // Wait for window adjust
          const sub = channel.windowAdjust$.subscribe(newSize => {
              if (newSize > 32768) {  // Enough window space
                  socket.resume()
                  sub.unsubscribe()
              }
          })
      }

      await channel.write(new Uint8Array(data.buffer, data.byteOffset, data.byteLength))
  })
```

  Benefits:

  1. Fixes Issue #10959 - Large file uploads will work reliably
  2. Prevents memory leaks - No unbounded buffer growth
  3. Matches ssh2 behavior - Proper flow control like the old implementation
  4. Enables high-throughput scenarios - SFTP, large data transfers, streaming

PR #10979, opened 2026-01-20
Ensure jump session is fully started before using it

balajieswaramoorthi:
When the device lost the and reconnects to network proxy jump is not working ...
image
Add check to wait for jump session to start and connect before attempting to use it for port forwarding. This prevents errors when the jump session hasn't been initialized yet.

Also improve jump host lookup logic in both sshTab component and sshMultiplexer service to handle various jump host identifier formats.
image


Eugeny:
Could you please outline the conditions for the issue happening at all? 

`setupOneSession` already does ensure the sesion is open as far as I see

Issue #10843, opened 2025-11-04
X Cannot read properties of undefined (reading 'filter')


wangka-na:
upgrade 1.0.225+ ，error：[X Cannot read properties of undefined (reading 'filter')]

now，Used 1.0.222，it's ok!

RichieB2B:
Duplicate of #10688

Issue #10730, opened 2025-09-09
Cannot read properties of undefined (reading 'filter')


jiyb:
Version: 1.0.227
Platform: macOS arm64 24.6.0
Plugins: quick-cmds, k8s, background
Frontend: xterm-webgl

RichieB2B:
I have the same issue with 1.0.228 under Windows 11.
Duplicate of #10688

ocpinfo:
I got the same error in 1.0229 MacOS

Issue #10688, opened 2025-08-27
SSH connection fails with "Cannot read properties of undefined (reading 'filter')" after upgrading to v1.0.224+


rejarevaldy:
**Describe the problem**:
After upgrading Tabby from `v1.0.223` to newer versions (`v1.0.224` and `v1.0.227`), SSH connections fail with the following error:

```
SSH  Connecting to OJT Jats
X  Cannot read properties of undefined (reading 'filter')
```

This issue does not occur in version `v1.0.223`.

**To Reproduce**:

1. Open Tabby `v1.0.224` or `v1.0.227`.
2. Select a saved SSH profile or enter an address to connect. 

<img width="466" height="133" alt="Image" src="https://github.com/user-attachments/assets/1cef0453-f722-4894-b712-a1254c3f5899" />

3. Attempt to connect.
4. Observe the error: `Cannot read properties of undefined (reading 'filter')`.

**Expected behavior**:
SSH should connect successfully, as it does on version `v1.0.223`.

**Actual behavior**:
SSH connection fails with `Cannot read properties of undefined (reading 'filter')`.

**Environment**:

* **Tabby version:** 1.0.224, 1.0.227 (bug occurs)
* **Last working version:** 1.0.223
* **OS:** Debian

frussane:
Have the same problem, we really could use a fix and seems like a top priority bug.
Any workaround other than using 1.0.223?

WanderingZombie:
In Windows 10 all SSH sessions fail to reconnect with "Cannot read properties of undefined (reading 'filter')", even after update to 1.0.228 today.

tomazz75:
I upgraded to 1.0.229 today and can't connect to any saved SSH session, same error "Cannot read properties of undefined (reading 'filter')". How come this isn't fixed?

EDIT: I can't connect to "old" sessions still opened, with "Reconnect" button.  If I close them and open new session then it works. 
@Eugeny wouldn't this be considered critical bug?

alexcool2016:
Me too, any idea?

 SSH  Connecting to app server
 SSH   !  Agent auth selected, but no running Agent process is found
 SSH   !  Agent auth selected, but no running Agent process is found
 X  Cannot read properties of undefined (reading 'filter')

Gelix:
Yeah, this bug is still here, I'm stuck on 1.0.223 on all my workstations.  Debug Console output
```
[...]

Issue #10682, opened 2025-08-26
connect fail


lunanting:
Version: 1.0.227
Platform: Windows x64 10.0.26100
Plugins: sync-config
Frontend: xterm-webgl

 SSH  Connecting to H618-demo-sd
 X  Cannot read properties of undefined (reading 'filter')

all connections  same

yLost:
Same here

FusionStreak:
I am getting the same error, after updating from 1.0.223 -> 1.0.227

System info:

```txt
Version: 1.0.227
Platform: Windows x64 10.0.26100
Plugins: quick-cmds, cloud-settings-sync
Frontend: xterm-webgl
```

From debug tools:

```txt
C:\Users\<USER>\AppData\Local\Programs\Tabby\resources\builtin-plugins\tabby-core\dist\index.js:15690 [hostApp] CLI arguments received: Object
error_handler.ts:45 ERROR TypeError: Cannot read properties of undefined (reading 'clear')
    at Object.next (C:\Users\<USER>\AppData\Local\Programs\Tabby\resources\builtin-plugins\tabby-terminal\dist\index.js:40664:47)
    at ConsumerObserver.next (Subscriber.ts:161:25)
    at Subscriber._next (Subscriber.ts:119:22)
    at Subscriber.next (Subscriber.ts:75:12)
    at emit (debounce.ts:84:20)
    at OperatorSubscriber._this._next (OperatorSubscriber.ts:70:13)
    at Subscriber.next (Subscriber.ts:75:12)
    at AsyncAction.<anonymous> (timer.ts:173:20)
    at AsyncAction._execute (AsyncAction.ts:120:12)
    at AsyncAction.execute (AsyncAction.ts:95:24)
    at AsyncScheduler.flush (AsyncScheduler.ts:40:27)
    at timer (zone.js:2355:41)
    at _ZoneDelegate.invokeTask (zone.js:402:31)
    at Object.onInvokeTask (ng_zone.ts:408:29)
    at _ZoneDelegate.invokeTask (zone.js:401:60)
    at Zone.runTask (zone.js:171:47)
    at invokeTask (zone.js:483:34)
    at ZoneTask.invoke (zone.js:472:48)
    at data.args.<computed> (zone.js:2335:32)
    at sentryWrapped (helpers.ts:87:17)
handleError @ error_handler.ts:45
// ....
[...]

PR #10994, opened 2026-01-24
fix(ssh): avoid undefined .filter() during SSH connection introduced …

### LINKED ISSUE #10619: X  Cannot read properties of undefined (reading 'filter')
macOS 15.6 (24G84)

 X  Cannot read properties of undefined (reading 'filter')

Gelix:
…in 1.0.224

SSHAlgorithmType.COMPRESSION was added in commit 020372c but is not filterable.
This loop missed the existing guard used elsewhere, resulting in .filter() being called on an undefined value.

potentially fixes
#10619 
#10635
#10672 
#10682
#10688 
#10730  
#10843

### DIFF
--- a/tabby-ssh/src/session/ssh.ts
+++ b/tabby-ssh/src/session/ssh.ts
@@ -352,9 +352,9 @@ export class SSHSession {
 
         const algorithms = {}
         for (const key of Object.values(SSHAlgorithmType)) {
-             if(key !== SSHAlgorithmType.COMPRESSION) { 
-            algorithms[key] = this.profile.options.algorithms![key].filter(x => supportedAlgorithms[key].includes(x))
-             }
+            if (key !== SSHAlgorithmType.COMPRESSION) {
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks for the PR? What are the conditions to trigger this bug actually? At this point, `this.profile` must already be a `ConfigProxy<SSHProfile>` with all fields defined

Gelix:
In all cases I observed, this issue does *not* occur on a fresh install or
clean configuration. It only happens when connections that were already
open prior to upgrading to 1.124 are restored after the update. When those
pre-existing connections are reused, the error appears.

The workaround is to close all existing connections and then reopen them.
This fix addresses it so the workaround is not needed.

Eugeny:
Thanks for invetigating this properly! I've fixed it in 10a4f54732909293a3e97c7cb252348ea6f6dd03 by ensuring recovered tabs also get a ConfigProxy as a `profile`

Eugeny:
@all-contributors add @Gelix for code

Issue #11006, opened 2026-02-01
The latest commit on the master branch is causing the tab to not display any content. (1.0.231-nightly.0)


ruanimal:
**Describe the problem**:
The latest [commit](https://github.com/Eugeny/tabby/commit/c7e6858d5bf0cfe0059aefb91a3f31112dae8876) on the master branch is causing the tab to not display any content.

This seems to be an issue related to the xterm version.

tag: 1.0.230 is fine

<img width="1762" height="946" alt="Image" src="https://github.com/user-attachments/assets/affda0ab-4448-4113-b25f-09559b0c2874" />

<img width="1317" height="1504" alt="Image" src="https://github.com/user-attachments/assets/e0e084b7-5f66-499f-8845-01fc7a299375" />

**To Reproduce**:
[Steps to reproduce the behavior]

build from soure, then open tabby

zeroleo12345:
same to me. It happend after merge these commit
<img width="2893" height="544" alt="Image" src="https://github.com/user-attachments/assets/467227b0-7f30-40ec-bdf6-b8320398e4cc" />

PR #11007, opened 2026-02-02
fix: upgraded webpack to fix ES6 class extension bug with xterm.js v6

ruanimal:
Upgraded webpack to 5.90.0+ (resolved to 5.104.1) in package.json

fix https://github.com/Eugeny/tabby/issues/11006


Eugeny:
Thank you!

Issue #10348, opened 2025-03-06
Pageant connection error


Figoware:
Posted to reopen [issue  10174](https://github.com/Eugeny/tabby/issues/10174)

```SSH  Connecting to XXXX
 SSH  Host key fingerprint:
 SSH   ecdsa-sha2-nistp256  SHA256:XXXXX 
 SSH   !  Failed to authenticate using agent: Error: Key(IO(Custom { kind: UnexpectedEof, error: "early eof" }))
```

Still can't connet with agent. 
Rollbak to 1.0.215 to get agent working.

SergioBondietti:
Yes, still have this issue also with the latest pageant 0.83

a15355447898a:
Similar question, but my logs are
```
 SSH  Connecting to nixos
 SSH  Host key fingerprint:
 SSH   ssh-rsa  SHA256:hPnpGkQ9V6DVQ6IjXvE4aQYHrQUZw5g1KbfgRhB4N8g 
 SSH   !  Failed to authenticate using agent: Error: IO(Os { code: 22, kind: InvalidInput, message: "无效的参数" })
```

Zit0X:
Hello
The version v1.0.230 fix the problem 🎉

docbert:
> Hello The version v1.0.230 fix the problem 🎉

Doesn't fix it for me.  Tabby 1.0.230 and Pagent 0.83 still gives the "early eof" error.

stom66:
Still getting early eof errors on v230. Sad to see it's been over a year since this was reported.

PR #11034, opened 2026-02-12
fix: improve agent authentication error handling and socket path validation

a15355447898a:
Fix this error: https://github.com/Eugeny/tabby/issues/10348#issuecomment-2808923436


Eugeny:
Thank you!

Issue #11040, opened 2026-02-18
Generate 256 color palette from 16 color palette specified.


tracker1:
Details specified in the linked post below, and already has been added to several terminals.

https://gist.github.com/jake-stewart/0a8ea46159a7da2c808e5be2177e1783

PR #11043, opened 2026-02-19
generate 256 palette

jake-stewart:
Hi Tabby team,

I believe that terminals should generate the extended 256-color palette based on the user's base16 theme.

The rationale and approach is written up [here](https://gist.github.com/jake-stewart/0a8ea46159a7da2c808e5be2177e1783).

This PR introduces a new `generate256ColorPalette` config option, enabled by default, which generates the colors 16-255 via interpolating the users base16 theme. No colors explicitly defined by the user are replaced.

The settings page is updated to allow toggling this functionality.


jake-stewart:
Closes #11040

jake-stewart:
This PR is not ready:

There is active [discussion](https://github.com/ghostty-org/ghostty/discussions/10852) at the Ghostty repo about approaching this slightly differently.

The idea is that light themes should not invert the palette so that ansi 16 remains black. This is with the intention of removing the breaking change of generating by default. Then, an opt-in option for "harmonious" colors which inverts the colors for light themes since there are advantages for doing so.

jake-stewart:
This PR is tested and ready.

It adds two configuration options that are **both opt-in**:

- `paletteGenerate` Whether to generate the extended 256-color palette based on the user's base16 theme.
- `paletteHarmonious` Whether to invert generated light themes for more semantic colors.

Both options have toggles in the color scheme settings tab. The harmonious toggle is only visible when palette generation is enabled.

The config and behaviour is consistent with other PRs/emulators.

----

The default 256-color palette interpolates from black to white. This means it works better for dark themes since the dark shades are closer to the background and light shades are closer to the foreground. For light themes, the opposite is true, and so a UI designed for dark themes using the default 256-color palette looks bad with light themes.

That is the purpose of the `paletteHarmonious` option: It inverts light themes so light/dark palettes are semantically equivalent. It makes shades interpolate from BG to FG instead of from black to white.

The `paletteHarmonious` is opt-in to be consistent with Ghostty. You could delete this option and only ever generate harmonious palettes. This would be fine since the palette generation is opt-in. However, you will be unable to make `paletteGenerate` default in the future without breaking changes.

Eugeny:
Thank you!

PR #11083, opened 2026-03-12
Fix frosted glass persistence

### LINKED ISSUE #11077: [Bug] "Frosted glass" effect does not persist after app restart
Description:
The "Frosted glass" setting (Appearance > Background) is lost every time the application is restarted, even though it was enabled in the previous session.

Reproduction Steps:

Go to Settings > Window > Appearance.

Enable "Frosted glass". (The effect applies correctly at this point).

Image

Close Tabby completely and restart the application.

Image

Actual Result: The frosted glass effect is gone, and I have to re-enable it in settings.

Expected Result: The frosted glass effect should remain active across restarts.

Environment:

Tabby version: 1.0.230

OS: 
Version: Windows 10 Home

Build Number: 22H2

Installation Date: 2022/4/11

OS Build Number: 19045.6466

uadhran:
## Summary
  The frosted glass toggle was saved through the debounced settings handler. If the app was closed quickly after toggling, the change could be lost. This makes the toggle save immediately
  instead.

  Fixes #11077.

  ## Changes
  - save `appearance.vibrancy` and `appearance.vibrancyType` directly via `config.save()`

  ## Testing
  Not run.

### DIFF
--- a/tabby-settings/src/components/windowSettingsTab.component.pug
+++ b/tabby-settings/src/components/windowSettingsTab.component.pug
@@ -40,7 +40,7 @@ h3.mb-3(translate) Window
 
     toggle(
         [(ngModel)]='config.store.appearance.vibrancy',
-        (ngModelChange)='saveConfiguration()'
+        (ngModelChange)='config.save()'
     )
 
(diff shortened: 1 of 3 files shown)


Eugeny:
Thank you!

PR #11089, opened 2026-03-14
fix: use switchMap instead of flatMap for plugin search to prevent st…

varunsub:
flatMap (mergeMap) does not cancel previous in-flight HTTP requests when a new search query arrives, causing race conditions where slower responses for old queries overwrite results for the current query.

switchMap cancels the previous inner observable on each new emission, ensuring only the latest search results are displayed.


Eugeny:
Thank you! @all-contributors add @varunsub for code

PR #11091, opened 2026-03-14
feat: add OS-specific config overrides

sourdusk:
## Summary
- Adds support for `windows`, `macos`, and `linux` top-level keys in `config.yaml` that override base config values when running on the corresponding platform
- Enables users who sync their config across machines to have per-platform hotkeys, fonts, profiles, etc. while sharing everything else
- Fixes a broken deep import in `tabby-ssh` (unrelated but discovered during build testing)

## How it works
At load time, the platform-specific sections are extracted and stashed. The current platform's section is deep-merged onto the base config. On save, all platform sections are reattached so they persist in the YAML file for syncing.

### Example config
```yaml
terminal:
  profile: "local:wsl"
  fontSize: 14

macos:
  terminal:
    profile: "local:default"
    font: "Menlo"

windows:
  terminal:
    profile: "local:wsl"

### DIFF
--- a/tabby-core/src/services/config.service.ts
+++ b/tabby-core/src/services/config.service.ts
@@ -336,15 +336,19 @@ export class ConfigService {
     }
 
     private extractPlatformSections (store: any): void {
-        const keys = ['windows', 'macos', 'linux'] as const
         this._platformSections = {}
-        for (const key of keys) {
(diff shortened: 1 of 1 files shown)


sourdusk:
This is ready for review at will.

Issue #11078, opened 2026-03-11
[Performance] Extreme UI degradation when loading 1,100+ profiles


leoberbert:
**Describe the problem:**

After migrating from SecureCRT, I imported approximately 1,100 SSH sessions (profiles) and their respective groups directly into Tabby's config.yaml file. After doing this, the application experiences severe performance degradation on Windows. The UI becomes extremely sluggish and unresponsive. Actions like opening the profile selector, navigating through the settings, or simply interacting with the terminal tabs take a long time to respond. It seems the app struggles to render or handle such a massive amount of profiles in the configuration file.

**To Reproduce:**

Export a large dataset of SSH sessions (around 1,100) from another terminal emulator (like SecureCRT) into a CSV file.

Use a custom Python script to parse this CSV and programmatically generate Tabby's config.yaml structure, appending all 1,100 sessions under the profiles array and organizing them into groups using UUIDs.

Replace the default config.yaml located in the Tabby configuration folder with this newly generated file.

Launch Tabby on Windows.

Attempt to open the profile selector menu or the settings page.

Observe the extreme UI lag, high resource consumption, and overall sluggishness.

Environment (please complete):

Tabby Version 1.0230]

OS: Windows 11

PR #11094, opened 2026-03-16
perf: fix UI degradation with large SSH config files. Issue #11078

hsdelirium:
Add mtime-based cache with Promise deduplication to OpenSSHImporter  to avoid re-parsing ~/.ssh/config on every change detection cycle.

To test, create ~1k fake hosts in ~/.ssh/config and measure the loading time of the profile list.

upd:
Performance: avoid redundant DOM updates — HTML binding directive now skips re-rendering when the value hasn't changed, reducing unnecessary DOM manipulation during change detection cycles.

Performance: profile descriptions computed once per refresh — profile descriptions are now cached after each profile list refresh instead of being recomputed on every render, which is significant when the list is large or descriptions are expensive to resolve.

Performance: profile list loaded once per refresh — the profiles settings page now fetches the full profile list once instead of twice when refreshing.

Performance: faster PowerShell detection on Windows — well-known installation paths are checked first before falling back to a full PATH scan, reducing startup latency on Windows.

Performance: SSH config parsing cached to disk — parsed SSH profiles are persisted to a local JSON cache keyed by the config file's modification time. On subsequent loads, if the file hasn't changed, the cache is used instead of re-parsing, speeding up startup when the SSH config is large.

Correctness: profile selector includes all profile fields — when building selector options from a profile, all profile fields are now forwarded instead of a manually maintained subset, preventing fields from being silently dropped as the profile schema evolves.

Correctness: added missing default field for terminal color scheme — the profile defaults object now explicitly includes terminalColorScheme, ensuring it is always present when profiles are merged.

### DIFF
--- a/tabby-core/src/directives/fastHtmlBind.directive.ts
+++ b/tabby-core/src/directives/fastHtmlBind.directive.ts
@@ -7,6 +7,7 @@ import { PlatformService } from '../api/platform'
 })
 export class FastHtmlBindDirective implements OnChanges {
     @Input() fastHtmlBind?: string
+    private _lastValue?: string
 
     constructor (
         private el: ElementRef,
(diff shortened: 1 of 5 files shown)


Eugeny:
Thank you! 

@all-contributors please add @hsdelirium for code

PR #11108, opened 2026-03-24
fix: hide blacklisted profiles from OS dock/taskbar menu

### LINKED ISSUE #10957: Build-in profiles shows on dock Menu even after hiding it in tabby setting
**Describe the problem**:
The built-in profiles still appear in the Dock menu of Tabby, even after they have been hidden via Tabby’s settings.

**To Reproduce**:
1. Open Tabby and go to the settings.
2. Hide the built-in profiles from the settings menu.
3. Right-click on the Tabby app icon in the Dock.
4. The hidden profiles still appear in the context menu despite the setting changes.

**Additional Info**:
OS: macOS 26 (Tahoe, latest version)
Tabby Version: latest

**Images**:

Image

---
Image

---
Image

uadhran:
## Description
Profiles that are hidden (blacklisted) in Tabby's settings should not appear in the OS-level quick-launch menus (Dock on macOS, Jump List on Windows). This PR ensures consistency between the internal profile selector and the external OS menus by applying the same profileBlacklist filter.

## Changes 
- Modified DockMenuService.update() in tabby/tabby-electron/src/services/dockMenu.service.ts to filter both regular and recent profiles against the configService.store.profileBlacklist.
- DockMenuService now stores ConfigService on the class, so profileBlacklist access compiles. 
- Pushed commit 57168b15 to fix/hide-hidden-profiles-dock.

## Fixes
Fixes #10957

### DIFF
--- a/tabby-electron/src/services/dockMenu.service.ts
+++ b/tabby-electron/src/services/dockMenu.service.ts
@@ -8,14 +8,14 @@ export class DockMenuService {
     appVersion: string
 
     private constructor (
-        config: ConfigService,
+        private configService: ConfigService,
         private electron: ElectronService,
         private hostApp: HostAppService,
         private zone: NgZone,
         private profilesService: ProfilesService,
         private translate: TranslateService,
     ) {
-        config.changed$.subscribe(() => this.update())
+        this.configService.changed$.subscribe(() => this.update())
     }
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you!

PR #11121, opened 2026-03-30
Add searchable vault feature for saved connections

xhatt:
**Tabby has been an excellent tool for managing my connections over the past few years.** 
One limitation I frequently encountered is that with many saved connections, viewing passwords manually is cumbersome, as the vault lacks sorting or filtering. To address this, I implemented a vault search feature last year, have been using it reliably for over six months, and now submit it for merge.
image


Eugeny:
Thanks!

PR #11144, opened 2026-04-09
Fix OSC 52 clipboard for strings > 1024 bytes

### LINKED ISSUE #6001: OSC 52 clipboard not support string length greater than 1024 bytes
**Describe the problem**:

OSC 52 clipboard not support string length greater than 1024/3*4.

**To Reproduce**:

run follow code in tmux
`tmux set-buffer -w "$( dd if=/dev/random bs=1024 count=1 | base64)"`

**Reason**:

  https://github.com/Eugeny/tabby/blob/45b6e21dbce0ef3155a71699c33e177e7ed2bf78/tabby-terminal/src/middleware/oscProcessing.ts#L15

the `data: Buffer` max length is 1024 bytes.  OSC string sequence will be splitted to mutiple data.  so the OSCPrefix and OSCSuffix will be in difference data.

ToastyTheBot:
Fixes #6001

### DIFF
--- a/tabby-terminal/src/middleware/oscProcessing.ts
+++ b/tabby-terminal/src/middleware/oscProcessing.ts
@@ -9,33 +9,62 @@ export class OSCProcessor extends SessionMiddleware {
     get cwdReported$ (): Observable<string> { return this.cwdReported }
 
     private cwdReported = new Subject<string>()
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks!

@all-contributors please add @ToastyTheBot for code

Issue #10892, opened 2025-11-18
The shortcut control panel system open automatically


lesslimes:
**Describe the problem**:
The shortcut control panel open automatically because of Tabby 

**To Reproduce**:
__Environment:__
Debian 12/13
__Graphical environment:__
KDE Plasma Wayland
__Tabby version:__
1.0.227 and above.

The problem occurs when the following actions are performed:
- Opening Tabby
- Editing a profile in ‘Profiles and Connections’
- When a profile is deleted
- When the Tabby settings page is opened and closed
- And randomly when I open a connection with the SSH quick connection menu.

DoctorCr4nk:
Hi, facing the same issue aswell:
Environment:
openSUSE Leap 15.6
Graphical environment:
KDE Plasma Wayland
Tabby version:
1.0.230-1

DeveloperBlue:
Experiencing this on 1.0.230 as well. KDE Plasma Wayland on Kubuntu.

The only way I was able to get it to stop was to launch tabby with `--ozone-platform=x11`

superm1:
Same on KDE/Wayland Ubuntu 24.04.

superm1:
However I don't see this on Arch at all.  I wonder if it's related to the (lack) of handling for global shortcuts in the older versions of Plasma.

If global shortcut capability can be made configurable that could be double checked.

PR #11157, opened 2026-04-16
Disable global shortcuts on older plasma versions

superm1:
At least for me this seems to avoid the shortcut panel from https://github.com/Eugeny/tabby/issues/10892 opening.


Eugeny:
Thanks! @all-contributors please add @superm1 for code

PR #11165, opened 2026-04-19
docs: add tabby-ssh-keymap to plugins list

mathys-lopinto:
## Add `tabby-ssh-keymap` to the plugins list

This PR adds [tabby-ssh-keymap](https://github.com/mathys-lopinto/tabby-ssh-keymap) to the community plugins section.

### What it does

Adds an indirection layer between Tabby SSH profiles and the physical paths of your private keys. Instead of hardcoding `file:///home/alice/.ssh/id_work` in every profile, profiles reference a logical name like `sshkey://work`, and each machine keeps a local, never-synced keymap file that resolves that name to the actual path.

### Why

Tabby's config sync is great, but SSH profiles with absolute key paths break the moment you sync to a machine with a different username or OS (`/home/alice/...` vs `C:\Users\Alice\...` vs `/Users/alice/...`). This plugin solves that without giving up config sync.

As a bonus, it also gives you a single source of truth for key paths locally — rotating a key or reorganizing `~/.ssh/` is a one-line edit in the keymap; every profile referencing that name follows automatically.

### Features

- One-click migration of existing `file://` entries in SSH profiles to `sshkey://`
- Cascade rename: renaming a key updates every profile that references it
- Key validation via `russh` parsing
- `~/` home expansion
- Search & per-key usage count in the settings UI
- Fully backward compatible — existing `file://` paths keep working

### Links

- npm: https://www.npmjs.com/package/tabby-ssh-keymap
- Repo: https://github.com/mathys-lopinto/tabby-ssh-keymap
- License: MIT


Eugeny:
Thanks for contributing to the ecosystem :)

PR #11166, opened 2026-04-19
Fix trailing semicolons in clickable URLs

### LINKED ISSUE #11134: Bug: Hyperlink parser includes trailing semicolon in clickable URLs
**Describe the problem**:
The hyperlink parser incorrectly includes a trailing semicolon (`;`) as part of clickable URLs. This happens when a URL is followed by a semicolon in the terminal output, which is a common occurrence in Postfix/SMTP logs. When clicking the link, the browser attempts to open the address with the semicolon included, leading to an invalid URL or 404 error.

- **Tabby version**: 1.0.230
- **OS**: Windows 11 x86

**To Reproduce**:
1. Output or paste a log line where a URL is followed by a semicolon delimiter. 
2. **Censored Example Log**:
   `Apr  6 09:39:55 mail postfix/smtpd[12667]: NOQUEUE: reject: RCPT from unknown[1.2.3.4]: 554 5.7.1 Service unavailable; Client host [1.2.3.4] blocked using dnsbl.spfbl.net; https://matrix.spfbl.net/1.2.3.4; from=<user@example.com> to=<target@example.com> proto=ESMTP helo=<[1.2.3.4]>`
3. Hover the mouse over the link `https://matrix.spfbl.net/1.2.3.4;`.
4. Notice that the semicolon is visually part of the link.
5. Click the link and observe the browser opening the URL with the trailing `;`.

lawrence3699:
Fixes #11134.

When a URL is followed by a log delimiter like `;`, Tabby currently includes that semicolon in the clickable range, so the browser opens the wrong URL.

This keeps internal semicolons intact and only prevents matches that end with `;`.

Validation:
- source regex repro for `https://matrix.spfbl.net/1.2.3.4;` now matches without the trailing `;`
- source regex repro still matches `https://example.com/foo;bar`
- source regex repro still matches `https://example.com/?a=1;2`
- built `tabby-linkifier` with webpack successfully


Eugeny:
Thanks!

PR #11170, opened 2026-04-21
feat(ssh): pre-populate vault password in auth prompts for MFA (fixes #9911)

### LINKED ISSUE #9911: Populate saved password when prompted for password
Hi,

We use YubiKey hardware tokens to provide MultiFactor Authentication. The integration is detailed here: https://lists.freeradius.org/pipermail/freeradius-users/2021-February/099521.html
 
This essentially means that we login as RSA style:
Username: joed
Password: secret<YubiKey TOTP press>
 
Feature request:
It would be great to be able to use the vault to prepopulate the saved password when prompted, allowing us to press the YubiKey hardware token to append the required TOTP.

syrexjohnson:
## Summary

- Pre-fills vault password in keyboard-interactive password prompts when the response field is empty, so users with hardware tokens (e.g. YubiKey) can append a TOTP to the pre-filled base password without re-typing it
- Pre-fills the prompt-password modal with the vault password for the same MFA workflow when SSH password auth is used

## Changes

**`tabby-ssh/src/components/keyboardInteractiveAuthPanel.component.ts`**
- Implements `OnInit` to load the vault password on component init
- For each prompt identified as a password prompt with an empty response, populates it from the vault
- Calls `ChangeDetectorRef.markForCheck()` to ensure the `OnPush` component re-renders with the pre-filled value

**`tabby-ssh/src/session/ssh.ts`**
- Before opening the `prompt-password` modal, loads the vault password and sets it as the initial `value` on the modal instance

## Use case (from issue #9911)

Users with RSA-style YubiKey MFA need to enter `<base-password><TOTP>`. With the vault storing just the base password, this change pre-fills the auth prompt so users only need to press their YubiKey token to append the TOTP and submit.

## Test plan

- [ ] Connect to an SSH server using keyboard-interactive auth with a vault-stored password — confirm the password field is pre-filled
- [ ] Press YubiKey / append TOTP to the pre-filled value and confirm authentication succeeds
- [ ] Confirm that prompts without a vault password still show an empty field
- [ ] Confirm that already-populated responses (from `method.savedPassword`) are not overwritten

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tabby-ssh/src/components/keyboardInteractiveAuthPanel.component.ts
+++ b/tabby-ssh/src/components/keyboardInteractiveAuthPanel.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, Output, EventEmitter, ViewChild, ElementRef, ChangeDetectionStrategy } from '@angular/core'
+import { Component, Input, Output, EventEmitter, ViewChild, ElementRef, ChangeDetectionStrategy, OnInit, ChangeDetectorRef } from '@angular/core'
 import { KeyboardInteractivePrompt } from '../session/ssh'
 import { SSHProfile } from '../api'
 import { PasswordStorageService } from '../services/passwordStorage.service'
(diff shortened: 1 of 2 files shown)


Eugeny:
Thanks!

PR #11193, opened 2026-04-28
fix(ssh): support OTP 2FA when auth mode is set to Password

MorseWayne:
Include keyboard-interactive auth method when auth is explicitly set to 'password'. Previously, keyboard-interactive was only included for 'auto' or 'keyboardInteractive' modes, causing OTP/2FA challenges to fail with 'Authentication rejected' when the user selected Password mode. This allows password auth to fall through to keyboard-interactive when the server requires a second factor.

当用户在 profile 设置中显式选择 "Password" 认证模式时，keyboard-interactive 也会被加入认证方法列表。这样当密码认证被服务器拒绝（因为需要 OTP），可以自动 fallthrough 到 keyboard-interactive 流程弹出 OTP 输入框，不再直接报 Authentication rejected

### DIFF
--- a/tabby-ssh/src/session/ssh.ts
+++ b/tabby-ssh/src/session/ssh.ts
@@ -280,7 +280,7 @@ export class SSHSession {
             }
         }
 
-        if (!this.profile.options.auth || this.profile.options.auth === 'keyboardInteractive') {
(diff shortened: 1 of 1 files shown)


Eugeny:
Superseded by #11492

Issue #11177, opened 2026-04-23
Release v1.0.231 — setup-x64.exe missing from assets


CryptoSiD:
Hi,

I was attempting to upgrade Tabby to v1.0.231, but it appears the Windows installer (setup-x64.exe) is missing from the release assets.

Is this intentional, or was it accidentally omitted from the release?

fabianod:
All Windows packages are missing!

iki:
Yes, [Release 1.0.231](https://github.com/Eugeny/tabby/releases/tag/v1.0.231) is missing binary release assets for Windows

⚠️ This breaks installing via binary package managers like [mise-en-place](https://mise.jdx.dev):

```sh
mise use -g github:eugeny/tabby
```

- **Possible solution**: Check release CI logs, fix and release properly

- **Temporary workaround**: Install previous release
    ```sh
    mise use -g github:eugeny/tabby@1.0.230
    ```

## Steps to Reproduce
1. check release assets on [1.0.231](https://github.com/Eugeny/tabby/releases/tag/v1.0.231)
2. check release assets on [1.0.230](https://github.com/Eugeny/tabby/releases/tag/v1.0.230)

## Expected Behavior

There should be binary release assets for Windows on both releases

## Actual Behavior

The binary release assets for Windows are missing on [release 1.0.231](https://github.com/Eugeny/tabby/releases/tag/v1.0.231)

Bonveio:
https://github.com/Eugeny/tabby/pull/11186#issuecomment-4339296803

`tabby-bin` from arch aur are also affected by this
https://aur.archlinux.org/packages/tabby-bin

<blockquote><img src="/static/images/favicon.ico" width="48" align="right"><div><strong><a href="https://aur.archlinux.org/packages/tabby-bin">AUR (en) - tabby-bin</a></strong></div></blockquote>

Eugeny:
Added the missing binaries

Issue #11176, opened 2026-04-23
Tabby v1.0.231 won't start


mzch:
**Describe the problem**:
v1.0.231 always crashes when starting.

**To Reproduce**:
1. Download release file.
2. Copy Tabby.app to /Applications folder/
3. Launch Tabby.app

** Spec**
Mac: Mac Studio Early 2022 (M1 Ultra)
OS: macOS 26.4.1
Tabby: v1.0.231

** Crash log **
```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------
Process:             Tabby [83704]
Path:                /Applications/Tabby.app/Contents/MacOS/Tabby
Identifier:          org.tabby
Version:             1.0.231 (1.0.231)
Code Type:           ARM-64 (Native)
Role:                Default
Parent Process:      launchd [1]
Coalition:           org.tabby [49448]
User ID:             501

Date/Time:           2026-04-23 14:48:25.0443 +0900
Launch Time:         2026-04-23 14:48:24.9754 +0900
Hardware Model:      Mac13,2
OS Version:          macOS 26.4.1 (25E253)
Release Type:        User

Crash Reporter Key:  4BD56778-F81D-DC1F-ACFF-C0FD059C49C9
Incident Identifier: C83B9F13-4606-458B-B10D-250C693BB8B8

Time Awake Since Boot: 360000 seconds

System Integrity Protection: enabled

Triggered by Thread: 0, Dispatch Queue: com.apple.main-thread

Exception Type:    EXC_BAD_ACCESS (SIGKILL (Code Signature Invalid))
Exception Subtype: UNKNOWN_0x32 at 0x00000001189c51a0
Exception Codes:   0x0000000000000032, 0x00000001189c51a0

Termination Reason:  Namespace CODESIGNING, Code 2, Invalid Page


VM Region Info: 0x1189c51a0 is in 0x10f790000-0x119448000;  bytes after start: 153309600  bytes before end: 11021919
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      Memory Tag 22               105e3c000-109e40000    [ 64.0M] rw-/rw- SM=PRV  
      GAP OF 0x5950000 BYTES
--->  __TEXT                      10f790000-119448000    [156.7M] r-x/rwx SM=COW  /Applications/Tabby.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework
[...]

Issue #11131, opened 2026-04-04
 Bug Report


likz0421:
Describe the bug
Tabby Version：1.0.230
OS Version: WindowsLTSC 11 24H2
After installing Tabby and trying to start it, it gets stuck on the startup logo screen and won't load the main interface - Local physical machine: Reinstalled the system and software, but the problem persists. Task Manager shows multiple Tabby processes, CPU usage is 0%, and there's no response
Additional context on troubleshooting attempts:
1. Tried adding `--disable-gpu` to disable hardware acceleration - didn't work
2. Deleted the `%APPDATA%\Tabby` configuration directory and reset settings - didn't work
3. Ran as an administrator - didn't work
4. Turned off antivirus software/firewall - didn't work
5. Updated graphics card driver to the latest version - didn't work

likz0421:
![image](https://github.com/user-attachments/assets/d7a13ec8-9a9e-4409-bd46-72313646285b)

vecinohk:
Have you tried installing the [Microsoft Visual C++ Redistributable](https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#latest-supported-redistributable-version) package?

itniels:
I just wanted to try Tabby as well, installed it on a fresh win11, and also stuck on the loading screen, the little progress bar quickly goes from 0 => 100 but then nothing. Clicking the Red close X also does nothing, you have to kill the process.

Update) the suggestion from @vecinohk works - but should maybe be a part of the installer instead? :)

Issue #10911, opened 2025-12-02
cannot start after upgraded to 1.0.228 / 1.0.229


nopeno:
**Describe the problem**:

i cannot start after upgraded. it showes as following:

```text
(node:2207502) [DEP0180] DeprecationWarning: fs.Stats constructor is deprecated.
(Use `tabby --trace-deprecation ...` to show where the warning was created)
kmsro: driver missing
[2207539:1202/115122.885100:ERROR:components/viz/service/main/viz_main_impl.cc:189] Exiting GPU process due to errors during initialization
kmsro: driver missing
[2207606:1202/115123.097258:ERROR:components/viz/service/main/viz_main_impl.cc:189] Exiting GPU process due to errors during initialization
[2207502:1202/115123.289487:ERROR:base/nix/mime_util_xdg.cc:139] Invalid mime.cache file does not contain null prior to ALIAS_LIST_OFFSET=44
Error: Bad window
    at ReadFixedRequest.callback (/opt/Tabby/resources/app.asar/node_modules/x11/lib/xcore.js:465:29)
    at ReadFixedRequest.execute (/opt/Tabby/resources/app.asar/node_modules/x11/lib/unpackstream.js:41:10)
    at UnpackStream.resume (/opt/Tabby/resources/app.asar/node_modules/x11/lib/unpackstream.js:165:30)
    at UnpackStream.write (/opt/Tabby/resources/app.asar/node_modules/x11/lib/unpackstream.js:102:10)
    at Socket.<anonymous> (/opt/Tabby/resources/app.asar/node_modules/x11/lib/xcore.js:88:21)
    at Socket.emit (node:events:519:28)
    at Socket.emit (node:domain:489:12)
    at addChunk (node:internal/streams/readable:561:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:512:3)
    at Socket.Readable.push (node:internal/streams/readable:392:5) {
  error: 3,
  seq: 9,
  badParam: 1,
  minorOpcode: 0,
  majorOpcode: 20
}
TypeError: Cannot read properties of undefined (reading 'type')
    at /opt/Tabby/resources/app.asar/node_modules/glasstron/src/native/linux_x11/x11.js:62:46
    at ReadFixedRequest.callback (/opt/Tabby/resources/app.asar/node_modules/x11/lib/xcore.js:490:39)
[...]

PR #11206, opened 2026-05-04
fix: show error page instead of hanging when plugin bootstrap fails

Fatty911:
## Summary

When Tabby's auto-updater produces a corrupted installation (e.g. `app.asar` updated to v1.0.231 but `builtin-plugins/` directory remains from v1.0.230), the renderer process fails to load plugins and **silently hangs at the logo screen**. Users see a frozen splash with no indication of what went wrong.

This PR adds a `showErrorPage()` function that replaces the preload-logo content with a visible error message when plugin bootstrap fails in both normal mode AND safe mode.

## The Problem

The auto-updater (Squirrel.Windows / electron-updater) can produce inconsistent installations where `app.asar` and `builtin-plugins/` are from different versions. This causes `nodeRequire()` to fail with "Cannot find module" errors during plugin loading.

In `app/src/entry.ts`, the existing error handling only logs to `console.error`:

```typescript
} catch (error2) {
    console.error('Bootstrap failed:', error2)
    // User sees: a frozen logo, forever
}
```

This is a recurring issue reported across dozens of issues:
- #9799, #9812, #9835, #9840, #9904, #10111, #10284, #10336, #10550, #10667, #10911, #11131, #11176, #11177

## The Fix

32-line addition to `app/src/entry.ts`:

1. **`showErrorPage()` function** — Replaces the `.preload-logo div` innerHTML with:
   - A clear "Startup Failed" heading
   - Explanation that the installation may be corrupted
   - The actual error message in a `<pre>` block for debugging
   - Instructions to reinstall from GitHub Releases (settings are preserved)
   - A hint to press `Ctrl+Shift+I` to open DevTools for detailed logs

2. **Call in the final catch block** — After both normal mode and safe mode bootstrap fail, `showErrorPage(error2)` is called instead of just `console.error`.

## Before / After

| Before | After |
|--------|-------|
| Logo hangs forever, no feedback | Visible error page with repair instructions |
| User must kill process via Task Manager | User knows exactly what to do |
| Requires reading debug logs to diagnose | Error details shown on screen |

## Screenshot

When the fix triggers, the splash screen is replaced with:

```
┌─────────────────────────────────────┐
│          Startup Failed             │
│                                     │
│  Tabby could not start because the  │
│  installation appears to be         │
│  corrupted. This typically happens  │
│  when an automatic update did not   │
│  complete properly.                 │
│                                     │
│  ┌───────────────────────────────┐  │
│  │ Error: Cannot find module     │  │
│  │ 'tabby-local/dist/index.js'   │  │
│  └───────────────────────────────┘  │
│                                     │
│  To fix this:                       │
│  1. Download from GitHub Releases   │
│  2. Run the installer to reinstall  │
│                                     │
│  Press Ctrl+Shift+I for DevTools    │
└─────────────────────────────────────┘
```

## Testing

The change is purely additive in an error path that was previously empty (only `console.error`). No existing behavior is changed.

To manually test: delete `resources/builtin-plugins/tabby-local/dist/index.js` (after backing up), then launch Tabby. The error page should appear instead of the frozen logo.

### DIFF
--- a/app/src/entry.ts
+++ b/app/src/entry.ts
@@ -54,6 +54,37 @@ async function bootstrap (bootstrapData: BootstrapData, plugins: PluginInfo[], s
     return moduleRef
 }
 
+function showErrorPage (error: any): void {
+    const el = document.querySelector('.preload-logo div') as HTMLElement
+    if (!el) return
+
+    const errorMessage = String(error?.message || error?.stack || error || 'Unknown error')
+    el.innerHTML = `
+        <div style="text-align: center; padding: 40px 20px; max-width: 520px;">
+            <h1 style="color: #e06c75; font-size: 22px; font-weight: 600; margin-bottom: 16px; font-family: -apple-system, sans-serif;">
+                Startup Failed
+            </h1>
+            <p style="color: #abb2bf; font-size: 14px; margin-bottom: 12px; line-height: 1.7; font-family: -apple-system, sans-serif;">
+                Tabby could not start because the installation appears to be corrupted.<br>
+                This typically happens when an automatic update did not complete properly.
+            </p>
+            <pre style="background: #1e222a; color: #e06c75; padding: 12px; border-radius: 6px;
+                font-size: 12px; text-align: left; overflow-x: auto; margin: 16px 0; white-space: pre-wrap;
+                word-break: break-all; max-height: 180px; font-family: 'SF Mono', 'Cascadia Code', monospace;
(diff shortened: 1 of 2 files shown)


Eugeny:
Builtin plugins are not packaged into app.asar.

Fatty911:
Closing this in favor of #11313. The maintainer feedback here clarified that builtin plugins are not packaged into app.asar, so the newer PR targets the actual bundled plugin load failure path and routes it through the existing safe-mode recovery instead of keeping this older error-page approach.

PR #11218, opened 2026-05-05
Add option to replace line breaks with spaces on paste

iiroak:
It is a simple addition because when there is no support for multi-line pasting in the shell and you use this for Claude Code or OpenCode, it becomes a problem to paste something with line breaks without it being sent as an "enter"
I don't really know if it's completely necessary, but it's useful in many cases so you don't have to be correcting it in a notepad

### DIFF
--- a/locale/app.pot
+++ b/locale/app.pot
@@ -2313,6 +2313,14 @@ msgstr ""
 msgid "Trim whitespace and newlines"
 msgstr ""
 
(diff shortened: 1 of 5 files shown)


Eugeny:
Thank you! However please note that you'll need to adjust translations on Crowdin as the ones in the repo will get automatically overwritten.

Eugeny:
@all-contributors add @iiroak for code

PR #11219, opened 2026-05-05
Fix themed backgrounds for side tabs and title bar

nguyenhuukhoi:
Fixes side tab and title bar backgrounds so they use theme colors instead of transparent/native window colors.

Changes:
- Use `--theme-bg-more-2` for vertical tab bars
- Apply themed background/text color to the custom title bar
- Let window controls inherit title bar color


Eugeny:
Thank you! @all-contributors please add @nguyenhuukhoi for code

PR #11250, opened 2026-05-07
fix: persist new SSH profiles from settings

QianYan-Art:
## Summary

This PR fixes a set of regressions in the profile editor that broke SSH profile creation/editing on Windows.

## Reproduction

1. Open Tabby on Windows.
2. Go to Settings -> Profiles & connections.
3. Create a new SSH profile or edit an existing one created from older config data.
4. Click Save.

Observed behavior before this change:

- The editor could throw `this.profile.options.algorithms[k] is not iterable` when legacy profiles contained `algorithms: {}`.
- The editor could throw `this.profile.__cleanup is not a function` when saving a newly created profile.
- After the crashes were avoided, the modal could still close without persisting the new profile, so the item never appeared in the list.

## Root cause

There were three separate issues in the save path:

1. `tabby-ssh/src/components/sshProfileSettings.component.ts` assumed every `profile.options.algorithms[k]` value was an iterable array.
2. `tabby-settings/src/components/editProfileModal.component.ts` unconditionally called `profile.__cleanup()` even when the edited object did not expose that helper.
3. The profile edit modal returned the stale original profile object instead of a serialized snapshot of the edited proxy-backed state, so newly created profiles were not persisted.

## Fix

- Normalize `profile.options.algorithms` during SSH profile editor initialization and tolerate legacy non-array values by treating them as empty algorithm selections.
- Guard the optional `profile.__cleanup()` call before invoking it.
- Return a serialized snapshot of the edited profile from the modal, so the caller receives the actual saved values for new profile creation.

## Validation

- Ran targeted `eslint` on:
  - `tabby-settings/src/components/editProfileModal.component.ts`
  - `tabby-ssh/src/components/sshProfileSettings.component.ts`
- Verified end-to-end locally on Windows:
  - create a new SSH profile
  - click Save
  - confirm the new item appears in the profile list

## Notes

Broader repository lint/typecheck currently reports unrelated baseline dependency/type issues in this environment, so validation for this change was kept focused on the touched files plus local end-to-end behavior.

### DIFF
--- a/tabby-settings/src/components/editProfileModal.component.ts
+++ b/tabby-settings/src/components/editProfileModal.component.ts
@@ -2,7 +2,7 @@
 import { Observable, OperatorFunction, debounceTime, map, distinctUntilChanged } from 'rxjs'
 import { Component, Input, ViewChild, ViewContainerRef, ComponentFactoryResolver, Injector } from '@angular/core'
 import { NgbActiveModal } from '@ng-bootstrap/ng-bootstrap'
-import { PartialProfileGroup, Profile, ProfileProvider, ProfileSettingsComponent, ProfilesService, TAB_COLORS, ProfileGroup, ConnectableProfileProvider, FullyDefined, ConfigProxy } from 'tabby-core'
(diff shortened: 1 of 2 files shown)


Eugeny:
Thank you for the PR, unfortunately it's incorrectly destroying the ConfigProxy instance through JSONification. I've already added the correct fixes in bca51380ea574ba01382ac3a8c5a2b63f9a586b8 and 690dc8f53ee05e6329bca1110c7329ebcc5ee9df

PR #11272, opened 2026-05-12
fix: improve menu readability and Chinese localization

mousehaozi:
Fix the file transfer menu becoming too transparent and hard to read when vibrancy mode is enabled. The menu now follows the vibrancy setting, enabling the blurred background effect only when vibrancy is turned on and keeping the normal menu appearance when it is disabled.
Also fix multiple Chinese localization issues, including missing translations, incorrect translations, and unnatural machine-translated wording.

Before Fix
e03389e97828d644fd772e96563c26de

After Fix
image


Eugeny:
Thank you for the PR but please see https://github.com/Eugeny/tabby/blob/master/locale/STOP.txt

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/Eugeny/tabby/blob/master/locale/STOP.txt">tabby/locale/STOP.txt at master · Eugeny/tabby</a></strong></div><div>A terminal for a more modern age. Contribute to Eugeny/tabby development by creating an account on GitHub.</div></blockquote>

PR #11275, opened 2026-05-12
fix: restore terminal redraw after tab reactivation

### LINKED ISSUE #11273: Inactive tab content disappears, requires minimize/maximize to redraw
When multiple tabs are open, switching away from a tab and then switching back shows a blank/empty tab — both the prompt and previous output are gone. The content only  reappears after minimizing and maximizing the window.
  
  ### Steps to reproduce
  
  1. Open Tabby with 2+ tabs
  2. Run some commands in Tab 1 (so there's visible output)
  3. Switch to Tab 2 and work there for a moment
  4. Switch back to Tab 1
  
  ### Expected behavior
  
  Tab 1 should display its previous content and prompt immediately.
  
  ### Actual behavior
  
  Tab 1 appears blank. Content only reappears after minimizing and maximizing the window (forcing a full repaint).
  
  ### Environment
  
  - Tabby version: 
  - OS: Linux
  - Rendering frontend: WebGL (default)
  

Version: 1.0.232
Platform: Windows x64 10.0.26200
Plugins: save-output
Frontend: xterm-webgl

uadhran:
## Summary
Fixes #11273 by restoring terminal redraw when switching back to a previously hidden tab.

## Root cause
Inactive terminal tabs zero their canvas resources after a delay to reduce GPU memory use. When a tab became visible again, the code used a narrow render-service path instead of the frontend's normal resize and refresh path. That left some tabs blank until a real window resize happened.

## Changes
- route terminal visibility transitions through a dedicated helper
- use the frontend's existing resize/redraw path when a tab becomes visible again
- keep the hidden-tab canvas cleanup behavior intact
- add a regression test for the visibility-transition contract
- fix the regression test import/build path so `build:typings` stays green
- switch the xterm overview ruler option back to the shape accepted by the current typings

## User impact
Previously, switching back to a terminal tab could show a blank tab until minimizing and maximizing the window. After this change, the terminal redraw path runs immediately when the tab becomes visible again.

## Validation
- `node --test --experimental-strip-types tabby-terminal/src/frontends/visibility.test.js`
- `yarn run build:typings`


alimbocl:
Hi Utkarsh, can you merge this into main? This will fix some major issues I've been having.

Eugeny:
Thanks!

Issue #11280, opened 2026-05-15
Warn when closing active conn (ssh) - no longer work in ver. 1.0.233


kubgrzeg:
**Describe the problem**:
Enabling "Warn when closing active connection" in ssh configuration no longer work when update from ver. 1.0.230 to ver. 1.0.233, The ssh connection closes without warning.
Tested on Windows 11 and Linux Mint Cinnamon 22.3.

cayenne17:
I've noticed the same thing on my end in Windows.
It seems to me that this has been the case since version 1.0.231.
diff v1.0.230=>v1.0.231 https://github.com/Eugeny/tabby/compare/v1.0.230...v1.0.231

@Eugeny : Could this be related to that commit?https://github.com/Eugeny/tabby/commit/438ac9a6ffa4c3e7d316f1253f596dea2b65b724

And editing the files [/tabby-ssh/src/components/sshTab.component.ts](https://github.com/Eugeny/tabby/compare/v1.0.230...v1.0.231#diff-105e5f5101ab3d725d6ce5cd6d86385c1bd5483071d539f60fa653d151ca7f1d)  +  [/tabby-ssh/src/components/sshProfileSettings.component.ts](https://github.com/Eugeny/tabby/compare/v1.0.230...v1.0.231#diff-e0ea4cc858c296c90833f0ce29a3ba1cd2c18d596f57bb69db4ab7880fcd047c)

When adding the feature in #2801, in the commit https://github.com/Eugeny/tabby/commit/4eefab5655478c2dbc3165bc249f8395eb015f4e, there were some modified files.

Aftab3008:
Can I work to fix it? I am interested. It is my first time to work on open source contribution.

PR #11285, opened 2026-05-17
Fix: Restore global "Warn when closing active connection" fallback

Aftab3008:
## Summary

* Restores the fallback check to the global `config.store.ssh.warnOnClose` configuration when determining if a warning dialog should be shown upon closing an active SSH connection tab.

## Problem

Active SSH connections close immediately without a warning dialog even if the global setting **"Warn when closing active connections"** is checked. This happens because:

1. `SSHTabComponent.canClose` only checks the profile-specific `this.profile.options.warnOnClose` property.
2. Since `warnOnClose` defaults to `null` for SSH profiles and is not exposed in the individual profile settings UI, the check `!this.profile.options.warnOnClose` always evaluates to `true`.
3. This immediately allows the tab to close, completely ignoring the global configuration check.

## Fix

Updates `canClose` in `sshTab.component.ts` to check `this.profile.options.warnOnClose ?? this.config.store.ssh.warnOnClose`. This allows the global warning preference to be respected when a profile-specific option is not set.

## Fixes

* **"Warn when closing active connection" in ssh configuration no longer work** #11280

## Test plan

- [x] TypeScript compiles with zero errors (`yarn run build`)
- [x] Verified active SSH connection tabs correctly show the disconnect warning dialog when the global setting is enabled

### DIFF
--- a/tabby-ssh/src/components/sshTab.component.ts
+++ b/tabby-ssh/src/components/sshTab.component.ts
@@ -198,7 +198,7 @@ export class SSHTabComponent extends ConnectableTerminalTabComponent<SSHProfile>
         if (!this.session?.open) {
             return true
         }
-        if (!this.profile.options.warnOnClose) {
+        if (!(this.profile.options.warnOnClose ?? this.config.store.ssh.warnOnClose)) {
             return true
         }
         return (await this.platform.showMessageBox(


Eugeny:
I'm afraid you're 5 minutes too late - but thank you for the PR!
6a0c0fd3b40ede6a9674d66514f3a7deee46afd7

Aftab3008:
Thank you for support. I will keep on working. I hope this time won't be late.

Issue #11281, opened 2026-05-15
Assigned Ctrl + Space hotkey does not execute action


xmha97:
## Description

In Tabby, I can assign the shortcut `Ctrl + Space` to commands in the keybinding settings without any issues.

However, after assigning it, the shortcut does not trigger the expected action when used in the terminal.

Other keyboard shortcuts work correctly.

---

## Steps to Reproduce

1. Open Tabby Settings
2. Navigate to Hotkeys
3. Assign any command to `Ctrl + Space`
4. Save the settings
5. Open a terminal session
6. Press `Ctrl + Space`

---

## Actual Behavior

* Pressing `Ctrl + Space` in the terminal does nothing

---

## Expected Behavior

* `Ctrl + Space` should trigger the assigned command like any other hotkey

---

## Environment

* OS: Linux
* Desktop Environment: KDE Neon (Plasma)
* Tabby version: AppImage build
* `Ctrl + Space` is not used as a system-wide or global shortcut
* No conflicts with KDE global shortcuts or other applications

Issue #2809, opened 2020-07-23
Terminus defaults to stealing global hotkey ctrl-space


SquidPony:
This is terrible behavior. I don't expect any app to by default make a new global hotkey on my system that overrides what open apps are doing.

Please change the default to not have any global hotkeys.

There are similar related issues, but the workaround of being able to disable a bad feature does not actually fix the problem. Especially because those issues are in a closed or stale status, finding why Terminus was doing this took a while to figure out.

https://github.com/Eugeny/terminus/issues/540
https://github.com/Eugeny/terminus/issues/1278
https://github.com/Eugeny/terminus/issues/998
https://github.com/Eugeny/terminus/issues/823
https://github.com/Eugeny/terminus/issues/683

SquidPony:
Found another closed issue related to this https://github.com/Eugeny/terminus/issues/1509

SquidPony:
Thanks for the fix! :D

Issue #683, opened 2019-02-21
Disable Ctrl-Space hotkey by default


JeroMiya:
The Ctrl-Space hotkey is supposed to toggle the terminal from any app. Problem is, that's also a hotkey for autocomplete in some editors. I didn't know why Terminus was gaining focus whenever I did it and so I had to reluctantly uninstall Terminus thinking it was a bug.

Later I found it's just a hotkey assignment in the preferences and I could disable it, but it might be a good idea to either disable this hotkey by default or ask on install or something like that, so new users (like me) don't think it's a bug.

Related to but not the same as: https://github.com/Eugeny/terminus/issues/540
I'm suggesting a change of default settings, 540 is suggesting easier way to change settings.

nuzar:
Please remove this hotkey.

Hammster:
Indeed, this hotkey should be changed. Or the user should be prompted with the option to deactivate it when happens the first time. To name a editor that is widely used which has difficulties due to ctrl+space is VScode.

TorbenJ:
Issue still exists, couldn't find any option to disable this. Everytime I activate intellisense in Visual Studio, Terminus pops up which is really annoying.

tehbeard:
Took a while to figure out why this was happening, and then longer to fix.

> **Edit:** So in setting up on my work computer I did see there is an option on the first run welcome screen for this. 
> My bad for missing it, I rushed thru to try out the SSH and profiles to see if it cut mustard as a replacement for putty (spoiler: it does, I really like how it can handle spinning up another terminal on the same ssh connection).
> That said I still hold that it being buried in with the other keybinds with no indication it's global is still an issue.
> I've tidied up what was an comment written in anger.

Please consider making it clear in the keybinds section that **toggle-window is global**, perhaps push it to the top of the keybinds so it can be easily found.

Nothing currently indicates it's a global keybind (beyond the first run welcome screen) and it's buried deep in the keybinds.

<img width="595" height="249" alt="Image" src="https://github.com/user-attachments/assets/b89821bd-d021-4cb0-9fd8-ca5ba0cfe9a9" />

Are there any other keybinds global that I need to worry about?

Issue #540, opened 2018-12-06
Make it easier to disable the global Ctrl+Space hotkey


jhorv-bvb:
I found [this closed issue](https://github.com/Eugeny/terminus/issues/139) and hunted everywhere in the settings for a button, finally realizing it should be listed the Config file | Defaults.  It was.  For anyone interested, you can disable this by adding the line

    toggle-window: null

under the `hotkeys:` section.

Eugeny:
It's actually right there on the Hotkeys page. You can disable or reassign it from there.

jhorv-bvb:
Huh!  Two things to report:

1. When I manually added the toggle-window: null line like I mention above, it did what I wanted while I was in Terminus, but the next time I closed/re-opened Terminus, it hung on the "splash screen" with the progress bar full, never fully starting.

2. There is a bug where it isn't actually visible in the Hotkeys tab.

![blankentries](https://user-images.githubusercontent.com/42846583/49610056-653b4480-f96b-11e8-870f-c22ed2a1e635.png)

But if I type "toggle" in the Hotkeys search bar, they are shown, and they are shown after that as well:

![nowshowing](https://user-images.githubusercontent.com/42846583/49610129-a895b300-f96b-11e8-8b10-ca2f0d0bca50.png)

Version info:

Version: 1.0.1
Platform: win32 10.0.17763
Plugins:

Eugeny:
What are all these empty lines in the table? Could you please post your complete config file?

jhorv-bvb:
I assume the empty lines are valid settings that aren't displaying properly.  Config attached.

[config.yaml.txt](https://github.com/Eugeny/terminus/files/2657532/config.yaml.txt)

jhorv-bvb:
I haven't seen this bug - empty lines in the Hotkeys table - in the past month of using Terminus.

Eugeny:
Thanks - let me know if you ever see it again and I'll reopen this.

h4sohail:
Encountering this again on clean install of Windows 11.

Eugeny:
Please capture DevTools console logs at the time of the issue: https://github.com/Eugeny/terminus/wiki/Debug-logs

PR #11287, opened 2026-05-18
fix: remove Ctrl+Space as default toggle-window global hotkey

uadhran:
## Problem

\`Ctrl+Space\` is currently the default global hotkey for \`toggle-window\` on all platforms. Because it is registered via Electron's \`globalShortcut\`, it is intercepted at the OS level before any key event reaches the renderer or xterm.

This causes two well-known issues:
1. Users who assign \`Ctrl+Space\` to a terminal-local hotkey see nothing happen — the global shortcut consumes the key first.
2. On Linux, \`Ctrl+Space\` is the standard IME toggle shortcut (ibus, fcitx), making it a poor default that conflicts with input methods.

Reported repeatedly: #540, #683, #2809, and most recently #11281.

## Fix

Remove \`Ctrl+Space\` from the platform defaults for \`toggle-window\`. Users who want a global toggle shortcut can assign one manually in Settings → Hotkeys.

## Test plan

- [ ] Verify \`toggle-window\` has no default hotkey after config reset
- [ ] Verify manually assigning a hotkey to \`toggle-window\` still works
- [ ] Verify \`Ctrl+Space\` assigned to a terminal-local hotkey now fires correctly
- [ ] Verify no regressions in \`new-window\` defaults (unchanged)

### DIFF
--- a/tabby-electron/src/config.ts
+++ b/tabby-electron/src/config.ts
@@ -5,19 +5,16 @@ export class ElectronConfigProvider extends ConfigProvider {
     platformDefaults = {
         [Platform.macOS]: {
             hotkeys: {
-                'toggle-window': ['Ctrl-Space'],
                 'new-window': ['⌘-N'],
             },
         },
         [Platform.Windows]: {
             hotkeys: {
-                'toggle-window': ['Ctrl-Space'],
                 'new-window': ['Ctrl-Shift-N'],
             },
         },
         [Platform.Linux]: {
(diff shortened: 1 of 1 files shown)


jruelle23:
+1 for this fix, I was going crazy with this shortcut not working in vscode before understanding what was happening.

Eugeny:
This shortcut is confirmed by the user on the first startup screen, it's not silently enabled.

PR #11291, opened 2026-05-18
fix: validate URL scheme before openExternal in linkifier

uadhran:
## Security Fix

Severity: Medium

The linkifier URLHandler and xterm built-in linkHandler called openExternal() with any URL matched from terminal output, without validating the scheme. The URL regex accepts any 3-9 letter scheme, so a malicious SSH server could print a file://, data:, or custom protocol URL and have it opened when the user clicks it.

A similar unchecked openExternal() call existed in fastHtmlBind for HTML-rendered links.

## Fix

Added an allowlist check (http, https, ftp, mailto, ssh) before invoking openExternal().

Files changed:
- tabby-linkifier/src/handlers.ts — URLHandler.handle()
- tabby-linkifier/src/decorator.ts — xterm linkHandler.activate
- tabby-core/src/directives/fastHtmlBind.directive.ts — rendered HTML link clicks

Note: IPHandler (prepends http://) and BaseFileHandler (prepends file://) are intentional and unchanged.

### DIFF
--- a/tabby-core/src/directives/fastHtmlBind.directive.ts
+++ b/tabby-core/src/directives/fastHtmlBind.directive.ts
@@ -20,9 +20,17 @@ export class FastHtmlBindDirective implements OnChanges {
         }
         this._lastValue = this.fastHtmlBind
         this.el.nativeElement.innerHTML = this.fastHtmlBind ?? ''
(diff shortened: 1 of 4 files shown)


Eugeny:
Already handled via safeExternalSchemes

Issue #11172, opened 2026-04-21
Feature request: Option to persist bell/activity indicator on tab until focused


michaelcm:
## Is your feature request related to a problem?

When a bell (`\a`) fires in a background tab, Tabby briefly flashes a blue dot indicator on the tab for about a second, then it disappears. If I'm not looking at the tab bar at that exact moment, I miss the notification entirely and have no way to know which tab had activity.

## Describe the solution you'd like

Add a setting to keep the bell/activity indicator visible on the tab until the user focuses that tab. This is how iTerm2 handles it — the blue dot persists until you click the tab, making it easy to see which tabs had activity even if you weren't watching.

Ideally this would be a configurable option, e.g.:
- `Bell indicator: Flash` (current behavior, default)
- `Bell indicator: Persist until focused`

## Describe alternatives you've considered

- **"Notify on activity"** from the tab context menu works but sends OS-level notifications, which is too noisy for frequent bells
- **Custom CSS** could potentially override the fade animation, but it's fragile and doesn't properly clear on focus

## Additional context

Other terminals that support persistent tab indicators:
- iTerm2: blue dot persists until tab is focused
- Kitty: configurable `tab_activity_symbol` that persists
- macOS Terminal.app: bell icon persists on tab until activated

PR #11292, opened 2026-05-18
fix: show persistent activity indicator on background tab when bell fires

uadhran:
## Problem

When a bell (`\a`) fires in a background tab, the tab header activity indicator (blue dot) briefly appears then disappears, making it easy to miss which tab had activity if you weren't watching at that exact moment. Related: #11172

## Root cause

The bell subscriber in `BaseTerminalTabComponent` handled visual/audible output but never called `displayActivity()`. The activity indicator was only triggered indirectly via `binaryOutput$` (which has a 1-second startup delay and can be affected by timing).

## Fix

Call `displayActivity()` inside the bell handler when the tab does not have focus. The indicator then persists on the tab header until the user clicks that tab, at which point `clearActivity()` is called automatically via the existing tab-switch logic — consistent with how other terminal output already marks activity.

## Behavior

- Bell fires in background tab → blue dot appears on tab header and stays visible
- User focuses that tab → indicator clears (existing behavior, no change)
- Bell fires in the currently focused tab → no indicator shown (tab is already in view)

### DIFF
--- a/tabby-terminal/src/api/baseTerminalTab.component.ts
+++ b/tabby-terminal/src/api/baseTerminalTab.component.ts
@@ -434,6 +434,9 @@ export class BaseTerminalTabComponent<P extends BaseTerminalProfile> extends Bas
             if (this.config.store.terminal.bell === 'audible') {
                 this.bellPlayer.play()
             }
+            if (!this.hasFocus) {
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you!

PR #11319, opened 2026-05-30
fix: ZMODEM transfer 

ruanimal:
# Fixes
##  1. transfer not terminated after rejecting the session
When the ZMODEM transfer confirmation dialog is rejected, the remote `rz`/`sz` process continues running and re-triggers `on_detect` repeatedly, causing the dialog to pop up in a loop.

## 2. error “Only thing after ZFIN” when cancel file dialog

image

## 3. wrong sz message order

image

## 4. sz message overwrited shell prompt
image


Eugeny:
Thanks!

PR #11340, opened 2026-06-10
fix(ssh): prevent keepalive flooding by mapping null/0 keepaliveInterval to undefined

### LINKED ISSUE #11302: SSH session closed frequently on existing profiles, but works fine on new ones
**Describe the problem**:

Version: 1.0.234
After rebooting my PC (Windows 10) and opening an existing tab (Tab 1) using an existing SSH profile (Profile 1) to connect to a device (Device 1), the SSH session closes multiple times within a single minute.
Trying to replicate the bug, I created a new SSH profile (Profile 2) to the same device (Device 1). Surprisingly, this new tab (Tab 2) from Profile 2 does not experience the session drop issue at all. Even more surprisingly, after doing this, the bug on Tab 1 was also fixed.
However, other existing tabs, such as Tab 3 using Profile 3 to connect to Device 2, still experience this frequent SSH session closed issue.

**To Reproduce**:
1. Reboot your PC.
2. Open Tabby and launch an existing SSH profile (Profile 1) to Device 1. The SSH session will close within one minute. Reconnecting does not fix it; the disconnects keep happening.
3. Create a new SSH profile (Profile 2) to Device 1 in a new tab (Tab 2). You will find that the SSH session no longer closes.
4. Check Tab 1 again, and the session closed issue is now gone for that profile as well.

linzack:
Hi,
this PR fixes #11302.
### Root Cause
An old profile (loaded from `localStorage.tabsRecovery`) can have `this.profile.options.keepaliveInterval` set to `null` (whereas a new profile defaults to `5000`).
In the current code:
`keepaliveIntervalSeconds: Math.round(this.profile.options.keepaliveInterval / 1000)`
Evaluating this with `null` yields `0` instead of `undefined` in javascript. 
Passing `0` will get `Some(Duration::from_secs(0))` in `russh`, which causes tabby to flood the ssh server with keepalive packets and leads to an immediate ssh `session closed`.
### Solution
This fix maps `null`, `undefined`, or `0` representing disabled keepalives to `undefined`, preventing packet flooding and resolving the disconnect loop.

### DIFF
--- a/tabby-ssh/src/session/ssh.ts
+++ b/tabby-ssh/src/session/ssh.ts
@@ -426,7 +426,7 @@ export class SSHSession {
                     key: this.profile.options.algorithms[SSHAlgorithmType.HOSTKEY].filter(x => supportedAlgorithms[SSHAlgorithmType.HOSTKEY].includes(x)),
                     compression: this.profile.options.algorithms[SSHAlgorithmType.COMPRESSION].filter(x => supportedAlgorithms[SSHAlgorithmType.COMPRESSION].includes(x)),
                 },
-                keepaliveIntervalSeconds: Math.round(this.profile.options.keepaliveInterval / 1000),
+                keepaliveIntervalSeconds: this.profile.options.keepaliveInterval ? Math.round(this.profile.options.keepaliveInterval / 1000) : undefined,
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks! @all-contributors add @linzack for code

Issue #11286, opened 2026-05-18
Black Tab Fix Regression


theAkito:
**Describe the problem**:
Since upgrade from `1.0.233` to `1.0.234` when opening `zellij`, the following gets pasted as input on the shell's prompt.

```
;25;rgb:0000/5f5f/afaf00/d7d7/5f5f\f5f/5f5f/87875f5f/d7d7/afaf0:8787/5f5f/d7d7111;rgb:8787/afaf/ffff135;rgb:afaf/5f5f/ffff161;rgb:d7d7/0000/5f5frgb:d7d7/5f5f/d7d7180;rgb:d7d7/afaf/8787201;rgb:ffff/0000/ffffrgb:ffff/8787/8787ffff/afaf/ffff/ffff/8787239;rgb:4e4e/4e4e/4e4e252;rgb:d0d0/d0d0/d0d0/dbdb;rgb:0000/5f5f/00007/afaf1;rgb:0000/d7d7/5f5fff/d7d760;rgb:5f5f/5f5f/8787787/ffff;79;rgb:5f5f/d7d7/afaf0000/0000rgb:8787/afaf/5f5f8787/d7d7/d7d7/0000/5f5ff/d7d75f155;rgb:afaf/ffff/5f5f171;rgb:d7d7/5f5f/ffff208;rgb:ffff/8787/0000226;rgb:ffff/ffff/0000237;rgb:3a3a/3a3a/3a3argb:9494/9494/9494eeee/eeee/eeee
```

The following appears, before the very first output from the shell, i.e. above the shell's banner.

```
^[\;25;rgb:0000/5f5f/afaf^[\00/d7d7/5f5f^[\\f5f/5f5f/8787^[\^[\5f5f/d7d7/afaf^[\0^[\:8787/5f5f/d7d7^[\111;rgb:8787/afaf/ffff^[\135;rgb:afaf/5f5f/ffff^[\161;rgb:d7d7/0000/5f5f^[\rgb:d7d7/5f5f/d7d7^[\180;rgb:d7d7/afaf/8787^[\201;rgb:ffff/0000/ffff^[\rgb:ffff/8787/8787^[\ffff/afaf/ffff^[\/ffff/8787^[\239;rgb:4e4e/4e4e/4e4e^[\252;rgb:d0d0/d0d0/d0d0^[\     __  ,
```

**To Reproduce**:
Update from `1.0.233` to `1.0.234`.

**Additional Information**:
This is the first time ever, this happens.

[...]

Issue #11212, opened 2026-05-05
Sometimes the console garbles into unintelligible noise


star7js:
Version: 1.0.230
Platform: macOS arm64 25.4.0
Plugins: background
Frontend: xterm-webgl
I notice this after using Claude Code with the terminal, but I'm not sure if it's related.  It generally takes a few hours to happen, and doesn't happen all of the time. When it does this, making it much bigger such as full screen can make it work again.

<img width="1698" height="2232" alt="Image" src="https://github.com/user-attachments/assets/5a0df7c2-20e6-4643-80a7-57cc056a1d5b" />

Issue #10303, opened 2025-02-13
All tabs went black


b-yond:
Version: 1.0.185
Platform: Windows x64 10.0.22631
Plugins: none
Frontend: xterm-webgl

I ran a sql query and the tab went black. I still had my cursor, but no text was displayed. I closed the tab, and realized all tabs were like this. Just reporting in case it helps.

Issue #10259, opened 2025-01-23
After a long period of inactivity, a blank black screen frequently appears, rendering it impossible to carry out any operations, forcing a manual exit.


Hello1M:
Version: 1.0.221
Platform: macOS arm64 24.2.0
Plugins: trzsz
Frontend: xterm-webgl

After a long period of inactivity, a blank black screen frequently appears, rendering it impossible to carry out any operations, forcing a manual exit.
在长时间不操作后，出现纯黑色界面，没办法进行任何操作，只能强制退出

beiliubei:
same issue.

Eugeny:
Please capture a debug log during the crash: https://github.com/Eugeny/tabby/wiki/Capturing-a-debug-log-from-the-SSH-backend

hackerpayne:
same issue

gabriel-n:
Same issue.
Version: 1.0.221
Platform: Linux Mint 22.1
Frontend: xterm-webgl

As asked before here my debug log : 

[155865:0213/135350.774077:ERROR:atom_cache.cc(230)] Add _NET_RESTACK_WINDOW to kAtomsToCache
[155900:0213/192657.824013:ERROR:shared_context_state.cc(1244)] SharedContextState context lost via ARB/EXT_robustness. Reset status = GL_UNKNOWN_CONTEXT_RESET_KHR
[155900:0213/192657.824167:ERROR:gpu_service_impl.cc(1164)] Exiting GPU process because some drivers can't recover from errors. GPU process will restart shortly.
[155865:0213/192657.845953:ERROR:gpu_process_host.cc(1007)] GPU process exited unexpectedly: exit_code=8704

point-source:
I am also having this issue after I return to tabby after inactivity. It doesn't always do it but it does do it frequently. The window will look fine and I can mouse over and click on it to activate the window but as soon as I type the first character, it goes blank and fully unresponsive. Even trying to quit does not work. I have to force quit (macOS). Here is a screenshot of how it looks:

![Image](https://github.com/user-attachments/assets/c85bafec-4095-40a5-ae5b-ce4bee8918cf)

HW: Macbook Pro  M4 Max (arm64)
OS: MacOS Sequoia 15.3.1 (24D70)

I will try to collect and edit this with debug logs next time it happens

jaloji:
same issue

amaymon:
Hi,
I could solve the problem, by swithing nvidia driver to nouveau.

[...]

PR #11354, opened 2026-06-13
fix(terminal): robust context loss recovery and direct DOM fallback during GPU resets

Gerry9000:
## Summary
When the operating system experiences a GPU driver crash/reset (e.g. Windows TDR `LiveKernelEvent 141`), Xterm.js's WebGL addon loses its context. Previously, Tabby's recovery mechanism would attempt to recreate the WebGL context, but background tabs (which are hidden with `offsetParent === null`) would immediately fail and prematurely exhaust their recovery retries. Additionally, falling back to the Canvas renderer during a severe GPU hang often results in permanent black screens since Canvas2D is also hardware-accelerated.

This PR overhauls `xtermFrontend` context recovery:
- **Visibility Throttling:** Defer recovery on background/hidden tabs until they are focused.
- **Dynamic Context Binding:** Added a `MutationObserver` to ensure context loss listeners bind dynamically when xterm reconstructs canvas nodes.
- **Direct DOM Fallback:** If WebGL fails to recover after 3 retries, the recovery now falls back directly to the native HTML DOM renderer, which is 100% immune to GPU crashes, guaranteeing the terminal remains visible.
- **Tab Reactivation Repaint:** Restored `renderService.clear()` and `handleResize()` when switching to background tabs to guarantee a fresh frame is painted.

## Verification
- Can be tested manually by initiating a GPU driver reset while Tabby is running in both foreground and background states. You can force a soft GPU driver reset in Windows using the keyboard shortcut `Win + Ctrl + Shift + B`.
- Re-tested against latest `master`.

**Example captured Crash Event (LiveKernelEvent 141)**:
```
Event Name: LiveKernelEvent
Problem signature:
P1: 141 (VIDEO_ENGINE_TIMEOUT_DETECTED)
P2: ffffb887c88f7010
P3: fffff8027f5295f0
```

## Related Issues
This PR is highly likely to address the following issues where users experience "black" or "blank" screens. Because these symptoms can occasionally stem from OS-specific rendering bugs, they are listed as related so reporters can confirm if this GPU crash recovery resolves their specific environment:

- **Context Loss & TDRs**: Often caused by the OS resetting the GPU driver under heavy load or idle suspend.
  - Related to #6296 (Explicit `CONTEXT_LOST_WEBGL` crash)
  - Related to #10259 (Black screens after long periods of inactivity, typical of Windows GPU power suspension)
  - Related to #10303 (All tabs went black)
  - Related to #8746 (Black tabs after full-screen transitions)
  - Related to #11286 (Black tab fix regression)
  - Related to #4653 (Blank screen issue)
  - Related to #8347 (Intermittent blank screens)
  - Related to #11212 (Garbled console noise; symptomatic of WebGL texture atlas corruption / silent rendering faults)

- **Sleep / Hibernation / Suspend**: When an OS suspends, the GPU context is typically destroyed. Upon waking, WebGL fires a context loss event. Our new `MutationObserver` and direct DOM fallback should allow Tabby to survive hibernation flawlessly.
  - Related to #7969 (Blank terminals after wake up from sleep)
  - Related to #3339 (Blank screen on Ubuntu after suspend)
  - Related to #5843 (Blank screen upon waking from hibernation/standby)


Eugeny:
Thanks!

PR #11368, opened 2026-06-18
fix(terminal): reset stale DEC modes on connectable reconnect

balajieswaramoorthi:
Connectable tabs (SSH and similar) reuse the same xterm.js frontend when a session ends and a new one is created. If the remote side had enabled mouse tracking (e.g. DECSET 1006 SGR coordinates) or bracketed paste (2004) and the connection dropped before a clean shutdown, those modes could remain active in the emulator. The user would then see mouse-driven input as literal text (e.g. coordinate fragments after right-click/paste).

After destroying the old session, call resetTerminalModes() so the parser applies DECRST for mouse modes and bracketed paste before the new session attaches. The base Frontend provides a no-op; XTermFrontend writes the standard escape sequences to xterm.

image


Eugeny:
Thanks! @all-contributors add @balajieswaramoorthi for code

PR #11324, opened 2026-06-03
feat: Add mouse wheel support for hotkeys

bluekani:
## Summary
Support wheel and middle-click hotkeys: WheelUp / WheelDown / WheelLeft / WheelRight / MiddleClick
This also enables zooming in and out of the device using the mouse wheel's hotkeys.

## Related issues
https://github.com/Eugeny/tabby/issues/9677
https://github.com/Eugeny/tabby/issues/10427

PR #11374, opened 2026-06-19
fix: Restore hotkey edit and remove actions

bluekani:
This fixes a regression introduced by the mouse wheel hotkey support changes.
The Edit and Remove actions for existing hotkeys can now be clicked again.

fix: #11324


Eugeny:
Thank you 🙏

PR #11323, opened 2026-06-02
feat(serial): Add custom baud rate input and improve SelectorModal options

bluekani:
## Summary
Serial: Enable user-defined custom baud rates.
https://github.com/Eugeny/tabby/issues/10808

## Changes

`tabby-core/src/api/selector.ts`
`tabby-core/src/components/selectorModal.component.ts`
- Added a new `freeInputPlacement` property to `SelectorOption`, allowing custom input fields to be placed at the top or bottom of the options list.
- Updated the `SelectorModalComponent` logic to honor the new `freeInputPlacement` property, ensuring custom input options appear in the desired position.

`tabby-serial/src/profiles.ts`
`tabby-serial/src/components/serialTab.component.ts`
- Added a feature to input a custom baud rate at the top of the baud rate selector.

PR #11382, opened 2026-06-23
feat(serial): Allow port selection on connect and fix temp baud rate handling

bluekani:
Depends on #11323

As many embedded developers are familiar with, the COM port assigned to a target device can change frequently.

This PR allows the Device field in serial profiles to be left empty, prompting the user to select a port each time the profile is launched.

It also fixes an issue where a baud rate entered in the select modal would unintentionally update a profile whose baud rate field was empty. The entered value is now used only for the current session and is no longer propagated back to the profile.

image

image

Tested on:
- Windows (x64)
- Ubuntu (WSL2)
- macOS (Intel)


Eugeny:
Already fixed

PR #11388, opened 2026-06-26
fix(terminal): restart visual bell shake on repeated BEL

### LINKED ISSUE #11303: visual bell CSS shake works a few times, then stops
**Describe the problem**:

The visual terminal bell shake is flaky when triggered repeatedly. With Terminal bell set to Visual, the first few BEL characters usually trigger the CSS shake, but after running printf '\a' multiple times, the shake sometimes stops  visibly restarting even though the terminal still receives bell events.

  This appears related to visualBell() resetting inline CSS animation from none back to the same terminalShakeFrames 0.3s ease value with setTimeout, which does not reliably restart a completed CSS animation on every BEL.

**To Reproduce**:

  1. Open Tabby.
  2. Go to Settings -> Terminal -> Sound.
  3. Set Terminal bell to Visual.
  4. Open a terminal tab.
  5. Run printf '\a' repeatedly, for example:

     printf '\a'
     printf '\a'                                                                                                                                                                                                                          
     printf '\a'
     printf '\a'
     printf '\a'

uadhran:
## Problem
Fixes #11303. With the terminal bell set to **Visual**, the shake animation stops firing after a few rapid bells. Once a shake is mid-flight, subsequent BELs no longer produce any visible shake.
## Cause
`visualBell()` cleared the animation (`animation = 'none'`) and restored it inside a zero-delay `setTimeout`. When bells arrive while a previous shake is still playing, the clear and the restore land in the same paint frame, so the browser never registers a change and skips the restart.
## Fix
Force a synchronous reflow (`void this.element.offsetWidth`) between clearing and reassigning the animation, so each bell reliably retriggers the shake.
## Testing
Reproduced the animation-restart mechanism in a headless browser: firing 10 bells at 60 ms intervals (each landing mid-shake), the old code restarted the animation only 2/10 times; the fix restarts 10/10.

### DIFF
--- a/tabby-terminal/src/frontends/xtermFrontend.ts
+++ b/tabby-terminal/src/frontends/xtermFrontend.ts
@@ -485,9 +485,12 @@ export class XTermFrontend extends Frontend {
     visualBell (): void {
         if (this.element) {
             this.element.style.animation = 'none'
-            setTimeout(() => {
-                this.element!.style.animation = 'terminalShakeFrames 0.3s ease'
-            })
+            // Force a synchronous reflow so the browser registers the cleared
+            // animation before it is reassigned. Without this, repeated bells
+            // arriving while a shake is still playing coalesce into a single
+            // frame and the animation fails to restart (#11303).
+            void this.element.offsetWidth
+            this.element.style.animation = 'terminalShakeFrames 0.3s ease'
         }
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks!

PR #11393, opened 2026-06-29
fix(ssh): prevent SFTP path navigation freeze

### LINKED ISSUE #11360: ~ symbol causes the software to freeze
**Describe the problem**:
[A clear and concise description of what the bug is.]

When entering the ~ symbol on the SFTP box to try to enter the home directory of the current user, the software freezes

**To Reproduce**:
[Steps to reproduce the behavior]

1. open SFTP box
2. double tap the path 
3. enter the ~ symbol

Image

A-Words:
## Summary

- stop SFTP breadcrumb generation when a relative path reaches `.`
- guard against fixed-point parent paths so invalid input cannot lock the renderer

## Root cause

The breadcrumb loop assumed every SFTP path was absolute. For `~`, `path.dirname()` reaches `.` and then keeps returning `.`, causing an infinite synchronous loop that freezes the UI.

Closes #11360

## Validation

- exercised path-segment regression cases for `~`, relative paths, absolute paths, `/`, and `.`
- transpiled the updated TypeScript component with esbuild
- ran `git diff --check`

### DIFF
--- a/tabby-ssh/src/components/sftpPanel.component.ts
+++ b/tabby-ssh/src/components/sftpPanel.component.ts
@@ -61,12 +61,16 @@ export class SFTPPanelComponent {
 
         let p = newPath
         this.pathSegments = []
-        while (p !== '/') {
+        while (p !== '/' && p !== '.') {
             this.pathSegments.unshift({
                 name: path.basename(p),
                 path: p,
             })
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks!

PR #11417, opened 2026-07-08
fix: various UI improvements

bluekani:
This PR contains several small UI and UX fixes:

- Open the color scheme tab matching the current colorSchemeMode instead of always opening the Light Mode tab.

- Fix the default color scheme being incorrectly shown as "Custom".

- Make the terminal toolbar transparent when using vibrancy.
image

- Prevent the terminal toolbar layout from breaking while dragging tabs.
image

- Add a close button to the restart notification banner.
image

- Disable dragging on navigation links in the Settings page.
image

### DIFF
--- a/tabby-core/src/theme.new.scss
+++ b/tabby-core/src/theme.new.scss
@@ -183,7 +183,7 @@ body {
 
 tab-body {
     terminal-toolbar {
-        background: var(--bs-body-bg);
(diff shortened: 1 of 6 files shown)


Eugeny:
Thanks :+1: - cherry-picked everything except the closeable banner which I disagree on

Issue #11400, opened 2026-07-02
Save Output.txt hotkey is not listed to be changed.


tdp4:
**Describe the problem**:
I discovered an unlisted hotkey Ctrl-Alt-Shift-5 that opens up a save output.txt dialog box.  I already use this as a system wide hotkey and whenever Tabby has focus it opened this dialog box at the same time as my other function activates.  Can this be listed to be changed or disabled?  Also found a Ctrl-Alt-Shift-2 that does something else too.

**To Reproduce**:
Be in a terminal session.  Press CTRL-ALT-SHIFT-5

bluekani:
Related: #5866

@Eugeny is [`tabby-terminal/src/features/debug.ts`](https://github.com/Eugeny/tabby/blob/master/tabby-terminal/src/features/debug.ts) still in use?

If not, would you be open to removing it, or would you prefer keeping it and extending it with additional debug options?

PR #11420, opened 2026-07-09
fix(terminal): expose debug hotkeys in settings for remap/disable

uadhran:
## Problem

Hidden debug shortcuts (`Ctrl-Shift-Alt-1` through `8`) were hardcoded in
`DebugDecorator` via a `keyup` listener. They did not appear in Settings →
Hotkeys, so users could not remap or disable them.

This caused real conflicts — e.g. `Ctrl-Shift-Alt-5` opens a save-`output.txt`
dialog while the same chord is used system-wide for another tool (#11400).

## Root cause

`tabby-terminal/src/features/debug.ts` matched modifier+number combinations
directly on the terminal element instead of going through `HotkeysService` and
the config-backed hotkey registry used by every other terminal action.

## Fix

- Register all eight debug actions in `TerminalHotkeyProvider` (e.g.
  `debug-save-output` → “Save session output (debug)”).
- Add config defaults preserving the existing chords (`Ctrl-Shift-Alt-1` … `8`).
- Replace the hardcoded `keyup` handler with a `hotkeys.hotkey$` subscription.

Users can now search “debug” in Hotkeys and clear or remap any binding.
Default behavior is unchanged for configs that have not been edited.

## Testing

- ESLint pass on changed files
- `tabby-terminal` webpack build pass
- Verified hotkey match logic: `Ctrl-Shift-Alt-5` → `debug-save-output`; empty binding disables match
- Electron dev app smoke start (`TABBY_DEV=1`)

Addresses #11400

### DIFF
--- a/tabby-terminal/src/config.ts
+++ b/tabby-terminal/src/config.ts
@@ -6,6 +6,14 @@ export class TerminalConfigProvider extends ConfigProvider {
     defaults = {
         hotkeys: {
             'copy-current-path': [],
+            'debug-save-state': ['Ctrl-Shift-Alt-1'],
+            'debug-load-state': ['Ctrl-Shift-Alt-2'],
+            'debug-copy-state': ['Ctrl-Shift-Alt-3'],
+            'debug-paste-state': ['Ctrl-Shift-Alt-4'],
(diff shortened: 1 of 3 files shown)


Eugeny:
Thanks!

Issue #11328, opened 2026-06-05
Feature Request: Option to show tab color on the index number (like Terminus)


leafgitgub:
In Terminus (predecessor of Tabby), when you assigned a custom color to a tab (e.g., via "Change tab color"), the color appeared as a rounded background on the **tab index number** (the `①` `②` `③` badge). This was a compact, elegant visual indicator — the color was right where your eyes naturally land.

In Tabby, the tab color is now shown as a thin `.colorbar` strip at the **bottom** of each tab header. It's less prominent, takes up vertical space, and doesn't match the Terminus aesthetic that many long-time users are accustomed to.

### Describe the solution you'd like

Add a configuration option (e.g., `terminal.tabColorLocation: "index" | "colorbar"`) that allows users to choose **where** the custom tab color is displayed:

- `"index"` — apply `tab.color` as the `background-color` of the `.index` element (the numbered badge), like Terminus used to do.
- `"colorbar"` (default) — keep the current bottom-strip behavior for backward compatibility.

### Relevant code pointers

The template is in `tabHeader.component.pug`:

```pug
.colorbar([style.background-color]='tab.color', *ngIf='tab.color != null')
// ...
.index(*ngIf='...') {{index + 1}}
```

The `.colorbar` is rendered separately from `.index`. Adding a branch like:

```pug
.colorbar(*ngIf='config.store.terminal.tabColorLocation !== "index" && tab.color != null', ...)
.index([style.background-color]='config.store.terminal.tabColorLocation === "index" ? tab.color : null', ...)
```

would allow toggling between the two modes.

### Describe alternatives you've considered

Custom CSS injection cannot achieve this because `tab.color` is an Angular component property not exposed as a CSS custom property, so CSS alone can't dynamically read per-tab colors.

### Additional context

- This is purely a UX preference — Terminus veterans miss the old badge-style color.
- Screenshot comparison would help illustrate, but the request is self-contained.

PR #11421, opened 2026-07-09
feat(terminal): add option to show tab color on index badge

uadhran:
## Problem

When a custom tab color is set, Tabby always shows it as a thin strip at
the bottom of the tab header. Long-time Terminus users preferred the old
style where the color appeared on the numbered tab index badge instead.

## Root cause

`tabHeader.component.pug` always renders `.colorbar` separately from
`.index`, with no setting to choose where the color is displayed.

## Fix

- Add `terminal.tabColorLocation` config (`"colorbar"` | `"index"`),
  defaulting to `"colorbar"` for backward compatibility.
- In `"index"` mode, apply `tab.color` as the index badge background and
  hide the bottom colorbar.
- Add a **Tab color indicator** control under Settings → Window → Tabs
  (hidden when tab index is disabled).
- Style the colored badge with white text and a light shadow for contrast.

## Testing

- `tabby-core` and `tabby-settings` webpack builds pass
- Manual test: set tab color, toggle Bottom strip / Index badge in settings

Addresses #11328


Eugeny:
This does not align with my UI vision and is better suited as a plugin

Issue #11363, opened 2026-06-18
Settings UI freezes on Hotkeys tab: `TypeError: Cannot read properties of undefined (reading 'toggle')`


XingHe501:
Version: 1.0.234
Platform: Windows x64 10.0.26200
Plugins: command-tips
Frontend: xterm-webgl

### Description

When navigating to the Hotkeys settings tab, a `TypeError` is thrown in the console. Once this error occurs, the settings UI becomes completely unresponsive. Clicking on the left-side navigation menu to switch to other settings tabs has no effect, and the view is permanently stuck on the Hotkeys interface.

It appears that an unhandled exception in `HotkeySettingsTabComponent.getHotkeys` is breaking the Angular change detection cycle, locking the UI state.

### Steps to reproduce

1. Open Tabby.
2. Open Settings.
3. Click on the "Hotkeys" tab on the left sidebar.
4. The UI freezes; attempting to click any other sidebar toggle/tab fails.
5. Open Developer Tools to observe the `TypeError`.

### Expected behavior

The Hotkeys tab should render its configuration list successfully, and navigating to other settings tabs via the left sidebar should continue to function normally.

### Actual behavior

The view gets stuck on the Hotkeys interface, and the following error is logged:

### Error details / logs

```text
error_handler.ts:45 
ERROR TypeError: Cannot read properties of undefined (reading 'toggle')    
    at HotkeySettingsTabComponent.getHotkeys (D:\Tabby\resources\b…st\index.js:8712:22)    
    at HotkeySettingsTabComponent_ng_container_8_div_1_Template (template.html:12:29)    
    at executeTemplate (shared.ts:496:5)    
    at refreshView (shared.ts:369:7)    
    at refreshEmbeddedViews (shared.ts:1563:9)    
    at refreshView (shared.ts:396:5)    
    at refreshEmbeddedViews (shared.ts:1563:9)    
    at refreshView (shared.ts:396:5)    
    at refreshComponent (shared.ts:1613:7)    
    at refreshChildComponents (shared.ts:109:5)    
    at refreshView (shared.ts:426:7)    
    at refreshEmbeddedViews (shared.ts:1563:9)    
    at refreshView (shared.ts:396:5)    
    at refreshComponent (shared.ts:1613:7)    
    at refreshChildComponents (shared.ts:109:5)    
    at refreshView (shared.ts:426:7)    
    at refreshContainsDirtyView (shared.ts:1636:11)    
[...]

PR #11422, opened 2026-07-09
fix(settings): guard hotkey config lookup for missing plugin paths

uadhran:
## Problem

Opening Settings → Hotkeys can freeze the entire settings UI with:

```
TypeError: Cannot read properties of undefined (reading 'toggle')
```

Once thrown, Angular change detection breaks and the sidebar stops responding.

## Root cause

`HotkeySettingsTabComponent.getHotkeys()` walks dotted hotkey ids
(e.g. from plugins like `command-tips`) through `config.store.hotkeys`
without checking that each path segment exists. A plugin-registered id
whose config branch was never created makes `ptr` undefined on the next
token access.

## Fix

- **`getHotkeys`**: return `[]` when a path segment is missing or the
  resolved value is not an array (covers `__nonStructural` placeholders).
- **`setHotkeys`**: create missing nested objects when the user assigns
  a binding to a hotkey that has no config entry yet.

## Testing

- ESLint pass on changed file
- `tabby-settings` webpack build pass
- Verified lookup logic: missing plugin paths return `[]`, existing keys unchanged

Addresses #11363

### DIFF
--- a/tabby-settings/src/components/hotkeySettingsTab.component.ts
+++ b/tabby-settings/src/components/hotkeySettingsTab.component.ts
@@ -35,17 +35,14 @@ export class HotkeySettingsTabComponent {
     }
 
     getHotkeys (id: string): Hotkey[] {
-        let ptr: unknown = this.config.store.hotkeys
-        for (const token of id.split(/\./g)) {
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks!

Issue #11397, opened 2026-07-01
Mouse tracking escape sequences leak as garbled text after nested SSH session disconnect


chalone0808:
## Description

  When a program like tmux enables mouse tracking inside a nested SSH session, and the inner connection drops abruptly, mouse clicks produce garbled escape sequence text (e.g., `0;90;14M`, `2;93;5M`) on the command line
  instead of normal click behavior.

  ## Steps to reproduce

  1. Set up a reverse SSH tunnel: `ssh -N -R 2230:localhost:22 user@hostA`
  2. In Tabby, open an SSH tab to `hostA`
  3. From the shell on `hostA`, SSH through the tunnel: `ssh localhost -p 2230`
  4. In the nested session on `hostB`, start `tmux`
  5. Terminate the reverse SSH tunnel (kill the process from step 1)
  6. The nested SSH connection drops and you are returned to the shell on `hostA`
  7. Left-clicking in the terminal produces garbled text like `0;90;14M` on the command line
  8. Right-clicking produces `2;93;5M` plus clipboard paste content
  9. The terminal window shakes (visual bell) on each click

  This also applies to any scenario where a program that enables mouse tracking (tmux, vim, etc.) dies without sending the disable sequences — for example, killing a process, network interruption, or session timeout.

  ## Root cause

  When tmux enables mouse tracking (via `\e[?1003h` and `\e[?1006h`), xterm.js intercepts mouse events and sends SGR mouse report sequences (e.g., `\e[<0;90;14M`) as terminal input. If tmux dies due to an abrupt connection
  loss, no program ever sends the disable sequences (`\e[?1003l`) to turn off mouse tracking. xterm.js remains in mouse tracking mode permanently.

  Subsequent mouse clicks generate escape sequences that are sent to the shell, which doesn't understand them. The shell rejects the unrecognized `\e[<` escape prefix (ringing the terminal bell) and echoes the remaining
  coordinate digits as literal text on the command line.

  Additionally, `xterm.modes.mouseTrackingMode` reports `'none'` even when mouse tracking is actively generating sequences, making the built-in xterm.js API unreliable for detecting this state.

  ## Proposed fix

  1. **Track mouse mode state via output parsing**: Parse DECSET (`\e[?...h`) and DECRST (`\e[?...l`) sequences in the output stream flowing from the remote to xterm.js. Maintain an accurate set of which mouse tracking
  modes (1000, 1002, 1003) are currently enabled by the remote.

[...]

PR #11423, opened 2026-07-09
fix(terminal): reset terminal modes on connectable session destroy

uadhran:
## Problem

When a connectable session (SSH/telnet/serial) ends abruptly, programs like tmux may have enabled mouse tracking without sending disable sequences. xterm.js can retain stale terminal modes until manually reset.

## Root cause

`resetTerminalModes()` only ran on manual reconnect. It was not called when a connectable session is destroyed on disconnect.

## Fix

Call `resetTerminalModes()` in `onSessionDestroyed` to clear stale mouse tracking and bracketed-paste state when the session ends.

**Limitation:** Does not cover nested SSH where the outer session stays open (#11397). Maintainer feedback: bell-based heuristics would break legitimate TUI bells.

Addresses #11397

### DIFF
--- a/tabby-terminal/src/api/connectableTerminalTab.component.ts
+++ b/tabby-terminal/src/api/connectableTerminalTab.component.ts
@@ -68,6 +68,7 @@ export abstract class ConnectableTerminalTabComponent<P extends ConnectableTermi
         super.onSessionDestroyed()
 
         if (this.frontend) {
+            this.frontend.resetTerminalModes()
             if (this.profile.behaviorOnSessionEnd === 'reconnect' && !this.isDisconnectedByHand) {
                 this.reconnect()
             } else if (this.profile.behaviorOnSessionEnd === 'keep' || !this.shouldTabBeDestroyedOnSessionClose()) {


Eugeny:
This breaks mouse mode whenever a TUI app rings the bell. The best I can merge is a resetTerminalModes in onSessionDestroyed - everything else is a heuristic that _will_ break

uadhran:
Stripped to your suggestion — bell/parser heuristic removed. Diff is now just `resetTerminalModes()` in `onSessionDestroyed`.

Agree the bell approach would break legitimate TUI bells.

Note: nested-SSH case from #11397 (outer session stays open) won't be covered; direct disconnect / session end will.

uadhran:
CI Lint is failing on `yarn install` (`keytar` rebuild → `Cannot find module 'string-width'`), same on current `master` since the "dep pins" commit (`ea051560`). Not related to this 1-line change — the previous run (Jul 9) was green.

Happy to re-run once master CI is fixed.

uadhran:
CI is green now. The diff is still just `resetTerminalModes()` in `onSessionDestroyed`, matching what you said you'd merge — no bell/parser heuristic.

Nested SSH (outer session stays open) is still out of scope, as noted earlier. Ready for another look when you have time.

PR #11486, opened 2026-07-15
fix(terminal): refresh profile defaults on reconnect #9834

### LINKED ISSUE #9834: Tabby is not respecting the "Clear terminal after connection" when it's disabled.
**Describe the problem**:

I have disabled it in `Settings -> Profiles -> Advanced` and in the `custom ssh profile` too but it keeps clearing after I connect. I'm using version 1.0.211.

**To Reproduce**:
Disable the "Clear terminal after connection" setting and reconnect to a session that has previous data in it.

AngKernel:
## Description
 Fixes #9834.

  `ConfigProxy` captures the merged profile defaults when a terminal tab is created. If the global or group profile defaults are changed while the tab remains open, reconnecting the tab continues to use the stale defaults.

  This is particularly visible with `clearServiceMessagesOnConnect`: when `false` matches the new inherited default, the explicit profile value is removed during config cleanup. The existing proxy then falls back to its
  original default value of `true`, causing the terminal to be cleared after reconnecting.

  This change:

  - Adds support for updating the defaults held by an existing `ConfigProxy`.
  - Recomputes profile defaults using the existing provider, global, and group precedence.
  - Refreshes the profile defaults before initializing a reconnecting session.
  - Keeps explicitly configured profile values higher priority than inherited defaults.
  - Applies the updated defaults only at the reconnect boundary, avoiding live changes to an active connection.

  ## Verification

  - ESLint passed for all modified files.
  - `yarn build:typings` passed for all plugins.
  - `yarn build` completed successfully.
  - `git diff --check` passed.

  Verified on Windows using the Release build.

  ## AI Usage

  * [ ] Fully vibe coded

### DIFF
--- a/tabby-core/src/services/config.service.ts
+++ b/tabby-core/src/services/config.service.ts
@@ -105,6 +105,15 @@ export class ConfigProxy<T extends AnyRec> {
             return deepClone(defaults[key])
         }
 
+        this.__setDefaults = (newDefaults: T) => { // eslint-disable-line @typescript-eslint/unbound-method
+            for (const key in defaults) {
+                if (isStructuralMember(defaults[key]) && isStructuralMember(newDefaults[key])) {
+                    (this as any)[key].__setDefaults(newDefaults[key])
+                }
+            }
+            defaults = newDefaults
+        }
+
         this.__setValue = (key: keyof T, value: any) => { // eslint-disable-line @typescript-eslint/unbound-method
             if (deepEqual(value, this.__getDefault(key))) {
                 // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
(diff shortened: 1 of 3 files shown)


AngKernel:
### Before
tabby-9834-before

  With `Clear terminal after connection` disabled, reconnecting the existing tab clears its terminal history.

  ### After
tabby-9834-after

  The profile defaults are refreshed before reconnecting, and the existing terminal history is preserved.

AngKernel:
### Test scenario

  Open an SSH session and generate some terminal output. Without closing the existing tab, disable Clear terminal after connection in both Settings → Profiles → Advanced and the corresponding custom SSH profile. Then return to the same
  tab and select Reconnect. Before this fix, the existing terminal output is cleared because the tab continues using stale profile defaults. After this fix, the output remains visible, confirming that the updated setting is applied
  when reconnecting an existing session.

PR #11494, opened 2026-07-20
Fix window title not syncing with renamed tabs

### LINKED ISSUE #11489: Synchronizing window title upon renaming tabs
**Describe the problem**:
Currently, when renaming a tab in the Tabby app, the new custom title is not in sync with the window title.
The tab title is only updated the first time you change it; if you then switch between tabs, it reverts to the default title.

**To Reproduce**:
1. Open an SSH connection
2. Open additional connections to the same host
3. Rename the tabs (e.g., Session 1, Session 2, Session 3)
4. Click through the different tabs
5. The old original title reappears under “Window.”

It seems that the title of the tab is no longer synchronized properly when switching tabs.

Image

System: macOS 26.5.2
Tabby: Version 1.0.234

H-TTTTT:
## Summary

When renaming a tab, the new custom title was reflected in the tab header but **not** in the window title. The window title reverted to the default/terminal title as soon as the user switched between tabs.

## Root cause

The window title was bound to the active tab's `title` field. For terminal tabs, the terminal continuously overwrites `title` with its own title sequence (current directory, running command, etc.), while the user's chosen name is stored in `customTitle`. The tab header correctly resolves this as `customTitle || title`, but the window title in `selectTab` used `tab.title` directly, so it always showed the underlying terminal title once a tab switch occurred.

Additionally, renaming the currently active tab did not re-sync the window title immediately, because `BaseTabComponent.setTitle` only emits `titleChange` when no `customTitle` is set.

## Changes

In `tabby-core/src/services/app.service.ts`:

- `selectTab` now resolves the window title as `customTitle || title`, matching the tab header.
- `renameTab` now re-syncs the window title immediately when the renamed tab is the active one.

This keeps the window title in sync with the tab header across renames and tab switches.

Fixes #11489

### DIFF
--- a/tabby-core/src/services/app.service.ts
+++ b/tabby-core/src/services/app.service.ts
@@ -238,7 +238,7 @@ export class AppService {
             this._activeTab?.emitFocused()
             this._activeTab?.emitVisibility(true)
         })
-        this.hostWindow.setTitle(this._activeTab?.title)
+        this.hostWindow.setTitle(this._activeTab?.customTitle || this._activeTab?.title)
     }
 
     getParentTab (tab: BaseTabComponent): SplitTabComponent|null {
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you!

PR #11524, opened 2026-07-27
fix(electron): guard window opacity and traffic-light IPC against closed windows

fxd0h:
## Description

`window-set-traffic-light-position` and `window-set-opacity` are registered with a bare `ipcMain.on`, so every `Window` instance responds to the message regardless of which window sent it, and without checking that its `BrowserWindow` still exists.

Once a window is closed (`this.window` is set to `null`) its handlers stay registered. The next time any window sends one of these messages, the closed window's handler runs and throws:

```
TypeError: Cannot read properties of null (reading 'setWindowButtonPosition')
TypeError: Cannot read properties of null (reading 'setOpacity')
```

This is easy to reproduce by opening and closing windows in quick succession.

Both handlers now go through the existing `on()` helper, which scopes them to the sending window (`e.sender === this.window.webContents`) like the neighbouring `window-set-progress-bar` handler, plus a null check via optional chaining. As a side effect this also stops a `set-opacity` / `set-traffic-light-position` sent by one window from being applied to every open window.

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/app/lib/window.ts
+++ b/app/lib/window.ts
@@ -466,12 +466,12 @@ export class Window {
         this.window.on('move', onBoundsChange)
         this.window.on('resize', onBoundsChange)
 
-        ipcMain.on('window-set-traffic-light-position', (_event, x, y) => {
-            this.window.setWindowButtonPosition({ x, y })
+        this.on('window-set-traffic-light-position', (_event, x, y) => {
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks!

PR #11525, opened 2026-07-27
Fix Windows-only plugin install failures via broken postinstall/prepare scripts

Blauschirm:
## Description

The Plugin Manager installs plugins by running `npm install` in the user's plugins directory. Several third-party plugins declare `tabby-core`/`tabby-ssh`/`tabby-terminal`/`tabby-settings` as real `dependencies` instead of `peerDependencies` (the correct pattern, since these are already part of the running Tabby install). When that happens, npm fetches and installs fresh copies of these builtin packages from the registry and runs their own lifecycle scripts — which is where the two bugs below surface. Both are Windows-only.

### `tabby-ssh`: missing `postinstall:win32`

`postinstall` uses `run-script-os` to pick a script for the current platform. There's a `postinstall:darwin:linux: exit` no-op, but no `postinstall:win32` entry. Git history shows a real `postinstall:win32` (an `xcopy` step for `pagent.exe`) existed until the commit that migrated the SSH backend from `ssh2` to `russh`, which correctly dropped the now-unneeded copy step but never added a win32 no-op to match the existing darwin/linux one. `run-script-os` errors out (rather than skipping silently) when no script matches the current OS, so `postinstall` now fails on every Windows install of `tabby-ssh` — not just nested ones.

Since the script is a no-op on every platform at this point, this PR removes `postinstall`/`postinstall:darwin:linux` entirely, along with the now-unused `run-script-os` dependency.

### `tabby-terminal`: `prepare` needs a devDependency-only tool

`prepare` runs `patch-package` to apply `patches/ansi-color+0.2.1.patch` and `patches/zmodem.js+0.1.10.patch` before webpack bundles those two packages into `dist/index.js`. `patch-package` is a `devDependency`, so npm skips installing it when `tabby-terminal` is pulled in as someone else's dependency — `prepare` then tries to run a command that isn't there, and the install fails.

The patches themselves are irrelevant to a nested install either way: `ansi-color`/`zmodem.js` are `devDependencies` (bundled at build time, not needed at runtime by consumers), and `patches/` isn't part of `"files"`, so it's never shipped. Rather than removing `prepare` outright (which would also skip patch application during this repo's own build — `scripts/install-deps.mjs` relies on it running via each plugin's own `yarn install`), this PR guards it: `"prepare": "patch-package || exit 0"`. Patches still apply normally when `patch-package` is present; the nested-install case becomes a harmless no-op instead of a failure.

### Testing

Native module builds aside, this doesn't require running the full app to verify — I confirmed both fixes at the level Tabby's Plugin Manager actually operates (`@npmcli/arborist`'s `reify()`, same as `app/lib/pluginManager.ts`):

- With the fix: installing a plugin that declares `tabby-ssh`/`tabby-terminal` as real dependencies (both as local `file:` installs of the built packages, and via a real published plugin that pulls in `tabby-terminal`) completes cleanly.
- With the fix reverted (`git stash`), the identical install reproduces the original failures exactly — `tabby-ssh`'s `postinstall` fails with `run-script-os` not found/not runnable, and `tabby-terminal`'s `prepare` fails trying to run `patch-package`.
- With the fix applied, `patch-package` still runs and applies both patches correctly when it's available (verified `ansi-color`/`zmodem.js` get patched and `tabby-terminal` builds via webpack without the pre-existing patch, which fails on `ansi-color`'s old octal escape syntax).

## AI Usage

* [ ] Fully vibe coded


Blauschirm:
This issue is still blocking me from installing certain plugins.

Eugeny:
Thanks!

PR #11538, opened 2026-07-29
fix(core): keep window title in sync with renamed tabs

uadhran:
## Problem

After renaming tabs, switching between them restores the original host/session title in the OS window title bar. Tab headers keep the custom name; the window does not.

## Root cause

- Tab headers use `customTitle || title`.
- The window used only `title`.
- Dynamic titles (OSC / SSH) still called `setTitle()`, which overwrote `title` even when `customTitle` was set (it only suppressed `titleChange$`).

## Fix

- `setTitle`: no-op while `customTitle` is set so dynamic updates cannot clobber a rename.
- `renameTab`: clear `customTitle` briefly so `setTitle` can write `title` and emit `titleChange$` for the active tab.
- `selectTab`: set the window title from `customTitle || title`, matching tab headers.

Addresses #11489

### DIFF
--- a/tabby-core/src/components/baseTab.component.ts
+++ b/tabby-core/src/components/baseTab.component.ts
@@ -112,10 +112,11 @@ export abstract class BaseTabComponent extends BaseComponent {
     }
 
     setTitle (title: string): void {
-        this.title = title
-        if (!this.customTitle) {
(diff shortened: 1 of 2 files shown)


Eugeny:
Closed in #11494

PR #11592, opened 2026-08-12
fix(local): only accept directories as the session CWD

### LINKED ISSUE #11584: Windows: a file path can be used as the session CWD, causing CreateProcess to fail with error 267 (ERROR_DIRECTORY)
### Environment

- OS: Windows 11, build 10.0.26200 (x64)
- Tabby version: 1.0.235
- Plugins: none (built-in only)
- `useConPTY`: true (default)

### Bug description

On Windows, a local terminal tab can permanently fail to start with:

```
uncaughtException: Error: Cannot create process, error code: 267
    at new WindowsPtyAgent (node_modules\node-pty\lib\windowsPtyAgent.js:65:36)
    at new WindowsTerminal (node_modules\node-pty\lib\windowsTerminal.js:51:24)
    at Object.spawn (node_modules\node-pty\lib\index.js:30:12)
    at new d (dist\webpack:\tabby\lib\pty.ts:96:37)
```

Windows error `267` is `ERROR_DIRECTORY` — *"The directory name is invalid."*

The root cause is that Tabby can assign a **file path** (not a directory) as the session's working directory, and every validation on the way to `CreateProcessW` only checks for *existence*, never for *being a directory*. The bad value is then persisted into `localStorage.tabsRecovery`, so the failure survives restarts and spreads to newly opened tabs.

### Root cause

**1. `guessWindowsCWD()` scrapes anything that looks like a path out of the terminal output**

`tabby-local/src/session.ts:9`

```ts
const windowsDirectoryRegex = /([a-zA-Z]:[^\:\[\]\?\"\<\>\|]+)/mi
```

`tabby-local/src/session.ts:230-235`

```ts
private guessWindowsCWD (data: string) {
    const match = windowsDirectoryRegex.exec(data)
    if (match) {
        this.guessedCWD = match[0]
    }
}
```

Any `X:\...`-looking substring printed by the shell becomes the "current directory". In practice this regularly captures **file** paths, e.g. echoing a command line that contains `D:\tools\7z.exe` or `--credential_file='C:\...\creds.json'`.

**2. `getWorkingDirectory()` validates with `fs.access()`, which a file passes**

`tabby-local/src/session.ts:219-227`

```ts
cwd = cwd || this.guessedCWD

try {
    await fs.access(cwd)
} catch {
    return null
}
return cwd
```

**3. The bad value is persisted into the tab recovery state**

`tabby-local/src/components/terminalTab.component.ts:78-89`

```ts
async getRecoveryToken (options?: GetRecoveryTokenOptions): Promise<any> {
    const cwd = this.session ? await this.session.getWorkingDirectory() : null
    return {
        type: 'app:local-tab',
        profile: {
            ...this.profile,
            options: {
                ...this.profile.options,
                cwd: cwd ?? this.profile.options.cwd,
```

**4. It also propagates to every new tab opened while such a tab is focused**

`tabby-local/src/profiles.ts:53-64`

```ts
if (!profile.options.cwd) {
    if (this.app.activeTab instanceof TerminalTabComponent && this.app.activeTab.session) {
        profile.options.cwd = await this.app.activeTab.session.getWorkingDirectory() ?? null
    }
```

**5. The two remaining guards use `existsSync()`, which a file also passes**

`tabby-local/src/services/terminal.service.ts:42-45`

```ts
if (cwd && !fsSync.existsSync(cwd)) {
    console.warn('Ignoring non-existent CWD:', cwd)
    cwd = null
}
```

`tabby-local/src/session.ts:90-95`

```ts
let cwd = options.cwd || process.env.HOME

if (!fsSync.existsSync(cwd!)) {
    console.warn('Ignoring non-existent CWD:', cwd)
    cwd = undefined
}
```

**6. `CreateProcessW` rejects it**

`node-pty/lib/windowsTerminal.js:43` → `var cwd = opt.cwd || process.cwd();`
`node-pty/lib/windowsPtyAgent.js:65` → `conptyNative.connect(this._pty, commandLine, cwd, env, ...)`, which calls `CreateProcessW(..., lpCurrentDirectory = cwd, ...)`.

### Evidence

Dumping the persisted recovery state from an affected profile (`%APPDATA%\tabby\Local Storage\leveldb`):

```
"cwd":"C:\\"                              exists=True  isDirectory=True
"cwd":"C:\\Users\\<user>"                 exists=True  isDirectory=True
"cwd":"D:\\Rime\\exports\\Hamster"        exists=True  isDirectory=True
"cwd":"D:\\Rime\\weasel-0.17.4\\7z.exe"   exists=True  isDirectory=False   <-- a file
```

I confirmed the failure mode by calling `CreateProcessW` directly with the exact command line Tabby uses for the `CMD (clink)` profile
(`cmd.exe /k <tabby>\resources\extras\clink\clink_x64.exe inject`), varying only `lpCurrentDirectory`:

| `lpCurrentDirectory` | Result |
| --- | --- |
| `D:\Tabby` (a real directory) | OK |
| `C:\Users\<user>` (a real directory) | OK |
| `NULL` | FAIL, err=123 `ERROR_INVALID_NAME` |
| `""` | FAIL, err=123 `ERROR_INVALID_NAME` |
| a deleted directory | **FAIL, err=267 `ERROR_DIRECTORY`** |
| **a file instead of a directory** | **FAIL, err=267 `ERROR_DIRECTORY`** |

So the command line itself is fine; only an invalid `cwd` produces 267.

### How to reproduce

1. On Windows, open a local terminal tab (CMD or PowerShell).
2. Print something containing an absolute path to an existing **file**, e.g.:
   ```
   echo C:\Windows\System32\cmd.exe
   ```
3. Open a new local tab while that tab is focused (`Ctrl-Shift-T`), or just restart Tabby so the tab is restored from `tabsRecovery`.
4. The tab fails to start with `Cannot create process, error code: 267` and is destroyed immediately.

Because step 3 re-saves the poisoned `cwd`, the tab stays broken across restarts until `localStorage.tabsRecovery` is cleared or an explicit working directory is set on the profile.

### Suggested fix

Replace the existence checks with a directory check. E.g. a shared helper:

```ts
async function isDirectory (p: string|null|undefined): Promise<boolean> {
    if (!p) return false
    try {
        return (await fs.stat(p)).isDirectory()
    } catch {
        return false
    }
}
```

and apply it at:

- `tabby-local/src/session.ts:223` — replace `await fs.access(cwd)` with an `isDirectory()` check
- `tabby-local/src/session.ts:92` — replace `fsSync.existsSync(cwd!)` with `fsSync.statSync(cwd!, { throwIfNoEntry: false })?.isDirectory()`
- `tabby-local/src/services/terminal.service.ts:42` — same as above
- `tabby-local/src/session.ts:230` `guessWindowsCWD()` — validate the guess is a directory before assigning it to `this.guessedCWD` (and consider trimming the match back to its parent directory when it points at a file)

Additionally, the spawn failure is currently an **uncaught exception** in the main process (`lib/pty.ts:96`). Catching it and surfacing a readable message in the tab (e.g. *"Invalid working directory: &lt;path&gt;"*) would make this far easier to diagnose than a raw Win32 error code.

### Workaround

Set an explicit **Working directory** on the affected profile. This makes `profiles.ts:53` skip the inheritance path and gives `session.ts:90` a known-good directory. Clearing `localStorage.tabsRecovery` is also required once, to drop the already-persisted bad value.

### Related issues

- #11178 — same error, same `local:cmd-clink` profile; diagnosed there as a ConPTY problem, but disabling ConPTY only masks it
- #3913 — log shows `cwd: "...\node_modules\babel-loader\lib\index.js"`, i.e. the same file-as-cwd situation
- #3427 — a comment from 2024-07-11 already dumped `tabsRecovery` and noted "the `cwd` value is a file path, not a directory", but the originating code path was never identified

I'm happy to put up a PR for this if the suggested direction looks right.

werner-balonius:
Fixes #11584 — a local tab permanently failing to start with `Cannot create process, error code: 267` (`ERROR_DIRECTORY`), surviving restarts and spreading to newly opened tabs.

## Root cause

`guessWindowsCWD()` scrapes anything that looks like an absolute path out of the terminal output:

```ts
const windowsDirectoryRegex = /([a-zA-Z]:[^\:\[\]\?\"\<\>\|]+)/mi
```

Any `X:\...`-looking substring printed by the shell becomes the "current directory", which in practice regularly captures **file** paths — e.g. an echoed command line containing `D:\tools\7z.exe` or `--credential_file='C:\...\creds.json'`.

Every validation between there and `CreateProcessW` only checks for *existence*, which a file passes:

| Location | Check |
| --- | --- |
| `tabby-local/src/session.ts:223` | `await fs.access(cwd)` |
| `tabby-local/src/session.ts:92` | `fsSync.existsSync(cwd)` |
| `tabby-local/src/services/terminal.service.ts:42` | `fsSync.existsSync(cwd)` |

The bad value is then persisted into `localStorage.tabsRecovery` by `getRecoveryToken()` and propagated to every new tab opened while such a tab is focused (`profiles.ts:53`), so the breakage survives restarts and spreads.

`CreateProcessW` rejects a non-directory `lpCurrentDirectory` with error 267. Per the analysis in #11584, varying only `lpCurrentDirectory` with the exact command line Tabby uses for the `CMD (clink)` profile:

| `lpCurrentDirectory` | Result |
| --- | --- |
| a real directory | OK |
| `NULL` or `""` | FAIL, err=123 `ERROR_INVALID_NAME` |
| a deleted directory | FAIL, err=267 `ERROR_DIRECTORY` |
| **a file** | **FAIL, err=267 `ERROR_DIRECTORY`** |

So the command line is fine; only the `cwd` produces 267. Full evidence, `tabsRecovery` dumps and repro steps are in #11584.

### 1. `fix(local): only accept directories as the session CWD` — the actual fix

- adds `isDirectory()` / `isDirectorySync()` in a new `tabby-local/src/util.ts`
- uses them at the three existence checks above
- `guessWindowsCWD()` now drops a guess that isn't a directory instead of storing it

I deliberately did **not** implement the "trim the match back to its parent directory when it points at a file" idea floated in the issue. If the shell echoes `D:\tools\7z.exe`, the real CWD is very unlikely to be `D:\tools`, so that would trade a loud failure for a silently wrong CWD. Keeping the previous valid guess is the conservative behaviour.

### 2. `fix(local): report shell spawn failures instead of crashing` — separable

node-pty throws synchronously when `CreateProcessW` fails, and the `pty:spawn` handler ran it uncaught in the main process, so a failed spawn surfaced as an `uncaughtException` carrying a bare Win32 error code and no hint about what was wrong. This catches it, hands the message back over IPC together with the working directory that was attempted, and prints it in the tab.

This changes the `pty:spawn` IPC return value from `id` to `{ id }` / `{ error }`; `tabby-electron/src/pty.ts` is the only consumer. It's a separate commit on purpose — happy to drop it if you'd rather keep this PR to the validation fix alone.

## Testing

CI on my fork (same `Package-Build` workflow): **Lint, macOS x64, macOS arm64, Linux x64 and Windows x64 all green.** Linux arm64 failed on an unrelated network flake fetching the Electron binary during `electron-builder` packaging (`Get ".../electron-v38.8.6-linux-arm64.zip": EOF`), after the webpack compile had already succeeded.

I have not yet exercised a packaged build against the original repro from #11584 — glad to do that if it would help.


werner-balonius:
The red build jobs here are a GitHub release-CDN flake, not this change. Two distinct signatures, both in dependency install / packaging, never in compilation:

```
node_modules/electron postinstall  ->  RequestError: socket hang up
electron-builder packaging         ->  Get ".../winCodeSign-2.6.0.7z": EOF
                                       Above command failed, retrying 3 more times   (all 4 retries failed)
```

In the Windows x64 run the 136 MB Electron archive itself downloaded fine (357 ms) and webpack completed; it died later fetching `winCodeSign`.

Across the two runs on this PR plus a run of the same workflow on my fork, every target has gone green at least once:

| Job | [fork run](https://github.com/werner-balonius/tabby/actions/runs/31633462017) | [PR run 1](https://github.com/Eugeny/tabby/actions/runs/31636591658) | [PR run 2](https://github.com/Eugeny/tabby/actions/runs/31638906970) |
| --- | --- | --- | --- |
| Lint | pass | pass | pass |
| macOS x64 | pass | pass | pass |
| macOS arm64 | pass | pass | pass |
| Linux x64 | pass | pass | flake |
| Linux arm64 | flake | pass | flake |
| Linux armhf | – | flake | pass |
| Windows x64 | **pass** | flake | flake |
| Windows arm64 | – | pass | flake |

Lint (`build:typings` + `eslint`) passed on all three runs, and Windows arm64 exercises the same `electron-builder` path as x64.

I've stopped re-triggering so as not to burn more CI on a transient upstream outage — happy to rebase or re-run whenever you'd like a clean board.

werner-balonius:
Update: I pushed an empty commit to re-trigger CI. All latest CodeQL and Package-Build checks passed. The earlier failures were transient dependency/CDN download issues, and no code changes were needed.

Eugeny:
Thanks!

Issue #5559, opened 2022-01-20
proxyjump


valmar:
Version: 1.0.170
Platform: macOS x64 21.2.0
Plugins: none
Frontend: xterm-webgl

In my ssh/config file, I have something like:

Host my.serverurl.com
    HostName my.internalname.com
    ProxyJump my.proxyname.com

However, if I try to connect to my.serverurl.com in a new tab I get the following error:

`X  my.serverul.com: jump host "openssh-config:my.proxyname.com" not found in your config`

Would it please be possible to support ProxyJump in tabby? Thank you and keep up the good work

valmar:
Sorry opened by mistake

valmar:
Some progress.

If I change the ssh/config file to include a separate Host section for the ProxyJump, like this:

```
Host my.proxyname.com
HostName my.proxyurl.com

Host my.serverurl.com
HostName my.internalname.com
ProxyJump my.proxyname.com
```

Then it attempts to connect. Notice that the ProxyJump name must still have the format of an URL: It cannot be, for example:

```
Host proxy
HostName my.proxyurl.com

Host my.serverurl.com
HostName my.internalname.com
ProxyJump proxy
```

Or Tabby will complain. Unfortunately, it is still not working completely. I don't use ssh-agent, so it should ask for my password twice: once for the proxy and once for the final machine. Tabby does this, but the second password always fails with the following error:

```
 SSH  Connecting to <hidden>
 SSH  Host key fingerprint:
 SSH   SHA256  <hidden> 
 SSH    Keyboard-interactive auth requested: 
 X  All configured authentication methods failed
 X  Not connected
```

Slowly getting there. I can't wait to solve this and make Tabby my daily driver!




However, since I don't use ssh-agent, ssh should ask me for two passwords, one for the proxy and one for the final machine.

lahma0:
To add to this, config entries that don't have the full hostname next to 'Host' don't even work. For example, the following won't even work (which is honestly a bit confounding):

```
Host home
    HostName home.myhost.com
```

The only way to make it work is to change it to:

`Host home.myhost.com`

Eugeny:
Fixed hostname parsing - but could you please double check that authentication works if you create both profiles manually in Tabby?

lahma0:
[...]

PR #11598, opened 2026-08-17
fix(ssh): import hosts without explicit HostName

VALLIS-NERIA:
## Summary

- import concrete `Host` entries even when they do not specify `HostName`
- use the `Host` alias as the connection hostname in that case
- keep an explicitly configured `HostName` unchanged, including case-insensitive directive spellings

## Why

OpenSSH defines the default value of `HostName` as the name given on the command line. A config such as:

```sshconfig
Host server.example.com
    User admin
```

is therefore a complete, connectable host definition.

Tabby's older importer followed this behavior, but the parser rewrite in #10043 started discarding every computed host configuration without an explicit `HostName`. This also prevents concrete hosts from inheriting settings from wildcard blocks unless they redundantly repeat their hostname. The same use case has been reported in #5559.

This change restores OpenSSH-compatible fallback behavior without changing wildcard filtering or overriding explicit hostnames.

## User impact

Concrete OpenSSH config entries without `HostName` now appear as built-in Tabby SSH profiles and connect using their `Host` alias. Existing entries with an explicit `HostName` behave as before.

## Validation

- `./node_modules/.bin/eslint --format json tabby-electron/src/sshImporters.ts`
- `./node_modules/.bin/tsc --project tabby-electron/tsconfig.json --noEmit`
- verified with `ssh -G` that OpenSSH resolves an omitted `HostName` to the host alias and preserves an explicit `HostName`
- verified both fallback and explicit-hostname cases with Tabby's locked `ssh-config` 5.0.1 parser


Eugeny:
Thanks! @all-contributors add @VALLIS-NERIA for code

PR #11611, opened 2026-08-22
Allow custom names for individual split panes

### LINKED ISSUE #11359: Allow custom names for individual split panes inside a tab
**Is your feature request related to a problem? Please describe.**
Yes.

When I use one Tabby tab with many split panes on a large monitor, it becomes easy to lose track of which pane is doing what. This is especially difficult when several panes run very similar terminal sessions or tools at the same time.

A common workflow for me is to keep one tab split into many panes (sometimes 6–8) and run multiple CLI sessions in parallel. For example, I often run multiple Claude CLI sessions at once. After some time, the default pane titles are not distinct enough, so it becomes hard to quickly identify each pane.

I would like to be able to assign a custom name to each individual split pane inside a tab.

Ideally, each pane could have its own optional label shown in the pane label area. If no custom name is set, the current default title behavior could remain unchanged.

A possible UX could be:
- focus a pane
- trigger an action such as `Rename active pane`
- enter a custom name
- show that custom name for that pane until cleared or changed

This could be exposed through the command palette, a keyboard shortcut, or another action-oriented UI that does not conflict with normal terminal mouse behavior.

Current alternatives are not very convenient:

- relying only on the default terminal/session title
- opening more top-level tabs instead of using split panes
- trying to encode context in the shell title manually
- rearranging panes to remember them by position

These work only partially, especially when many panes are running similar sessions.

This would be especially useful for users who use split panes as a workspace layout.

I am not asking for a full workspace management feature — only a lightweight way to manually label panes inside one split tab.

If this sounds like a good fit for Tabby, I would be happy to try implementing it and opening a PR.

liewstar:
## Summary
- add a `Rename active pane` command and configurable `rename-pane` hotkey
- show custom pane names as compact persistent labels while preserving the existing rearrange overlay
- keep custom names across tab recovery and split layout duplication, and restore the dynamic title immediately when cleared

## Testing
- `yarn eslint -f json` on all changed TypeScript files
- `yarn tsc -p tabby-core/tsconfig.json --noEmit`
- `yarn webpack --config tabby-core/webpack.config.mjs`
- parsed Windows, macOS, and Linux default config YAML with `js-yaml`

Closes #11359


liewstar:
Closing this version to rework the pane rename UX around the pane's own context menu. A revised implementation will be tested locally before resubmission.

PR #11614, opened 2026-08-23
fix(local): restore mouse scrolling in Windows TUIs

### LINKED ISSUE #11051: mouse scroll impossible when use OpenCode TUI
platform：
windows 10 x86
tabby version 1.0.230

i was using tabby to launch OpenCode TUI ,but something strange happend, i could not scroll up with my mouse scroll button as using native cmd teminal to see what history text opencode has printed，however i could use pageup or pagedown key to scroll。

this is annoying，because i was in the habit of using mouse scroll to browse history。
what can i do to solve this issue？

liewstar:
## Summary

- use node-pty's bundled ConPTY implementation for Windows local sessions
- restore VT mouse mode and SGR mouse input passthrough for fullscreen TUIs
- copy and unpack the bundled ConPTY runtime files during Windows builds

## Root cause

The system ConPTY consumes `DECSET 1000/1002/1003/1006` output instead of forwarding it to xterm.js. xterm.js therefore never enables mouse reporting. It also turns SGR mouse input into Win32 mouse records, while OpenCode/OpenTUI reads VT input from stdin.

node-pty's bundled ConPTY preserves both sides of the VT mouse protocol. The build also needs `conpty.dll` and `OpenConsole.exe` next to `conpty.node`, outside ASAR.

## Validation

- `npx eslint tabby-local/src/session.ts`
- `node --check scripts/build-native.mjs`
- `npx webpack --progress --color` in `tabby-local`
- byte-level ConPTY probe: system ConPTY drops all four mouse modes; bundled ConPTY preserves all four and passes SGR wheel input through
- packaged Windows preview: xterm entered mouse-reporting mode and the PTY received both wheel-down (`CSI <65...M`) and wheel-up (`CSI <64...M`) reports
- OpenCode 1.18.21 smoke test in the packaged preview with mouse reporting enabled

Fixes #11051


ebifrier:
I tested this PR and can confirm it works.

In my case, mouse scrolling wasn't working when running Claude Code in fullscreen mode inside WSL, but after applying this patch it works as expected.

Environment:
- Tabby: built from master with this patch applied
- OS: Windows 10
- WSL: Ubuntu 26.04

Great PR — thank you for this fix!

chrisrude:
confirmed this fixes the issue with fullscreen claude code.  you can get a windows binary before it's merged in the "checks" section

Issue #11593, opened 2026-08-13
Tab recovery: closeWindow() disables recovery before the final saveTabs(), making it a no-op


marcodiniz:
**Version:** present on `master`; identical code in the installed 1.0.234 build
**Platform:** all (found on Windows 11)

`AppService.closeWindow()` clears the recovery flag and *then* awaits the final save:

```ts
// tabby-core/src/services/app.service.ts
async closeWindow (): Promise<void> {
    this.tabRecovery.enabled = false
    await this.tabRecovery.saveTabs(this.tabs)
    if (await this.closeAllTabs()) {
        this.hostWindow.close()
    } else {
        this.tabRecovery.enabled = true
    }
}
```

But `saveTabs()` returns early on exactly that flag:

```ts
// tabby-core/src/services/tabRecovery.service.ts
async saveTabs (tabs: BaseTabComponent[]): Promise<void> {
    if (!this.enabled || !this.config.store.recoverTabs) {
        return
    }
    window.localStorage.tabsRecovery = JSON.stringify(/* ... */)
}
```

So the save that is meant to capture final tab state on a clean close never writes anything. The newest state that ever reaches `localStorage.tabsRecovery` is whatever the debounced/interval save last wrote:

```ts
setInterval(() => {
    this.recoveryStateChangedHint.next()
}, 30000)

this.recoveryStateChangedHint.pipe(debounceTime(1000)).subscribe(() => {
    this.tabRecovery.saveTabs(this.tabs)
})
```

`tabsChanged$` covers open/close/reorder, so those are saved within ~1s. Everything else — scrollback growth, working directory changes, a fresh `restoreFromPTYID` — only lands on the blind 30s tick. Closing the window inside that window silently discards it.

### Repro

Derived from reading the code path on `master` rather than executed against a build, so please treat the timings as predicted rather than measured:

1. `recoverTabs: true` (default).
2. Open a local terminal tab and let it settle so a save has happened.
3. `cd` into a distinctive directory.
4. Close the window within 30 seconds.
5. Reopen Tabby — the restored tab's working directory is the pre-`cd` one, and its restored scrollback ends before step 3.

Waiting 30+ seconds before closing at step 4 should restore correctly, which would isolate this to the missed final save rather than to serialization.

### Suggested fix

Save first, then latch the flag off — this keeps the original intent (no further saves during teardown) while making the final save effective:

```ts
async closeWindow (): Promise<void> {
[...]

PR #11621, opened 2026-08-24
fix(core): save tab recovery state before disabling it on close

uadhran:
Addresses #11593

## Problem

`AppService.closeWindow()` set `tabRecovery.enabled = false` and then called `saveTabs()`. `saveTabs()` returns immediately when that flag is off, so the snapshot that is supposed to run on a clean close never wrote. Recovery on the next launch was whatever the last 1s/30s debounce had stored.

`enabled = false` is still required *during* teardown: `closeAllTabs()` destroys tabs and that would otherwise save an empty list over the snapshot.

## Fix

Save first, then latch the flag off:

```ts
async closeWindow (): Promise<void> {
    await this.tabRecovery.saveTabs(this.tabs)
    this.tabRecovery.enabled = false
    if (await this.closeAllTabs()) {
        this.hostWindow.close()
    } else {
        this.tabRecovery.enabled = true
    }
}
```

`saveTabs()` is unchanged. If a tab blocks close, recovery is turned back on as before.

## Verification

Ran the real `saveTabs` / `closeWindow` bodies:

- save while enabled writes
- save while disabled is a no-op
- close writes current tab state, then disables
- a follow-up save of `[]` does not wipe that snapshot
- if a tab blocks close, recovery is re-enabled and the window stays open

`eslint` on `app.service.ts` is clean.

### DIFF
--- a/tabby-core/src/services/app.service.ts
+++ b/tabby-core/src/services/app.service.ts
@@ -472,8 +472,8 @@ export class AppService {
     }
 
     async closeWindow (): Promise<void> {
-        this.tabRecovery.enabled = false
         await this.tabRecovery.saveTabs(this.tabs)
+        this.tabRecovery.enabled = false
         if (await this.closeAllTabs()) {
             this.hostWindow.close()
         } else {


Eugeny:
Thanks!

PR #11625, opened 2026-08-27
fix(local): resolve guest cwd against the shell's fsBase before validating

### LINKED ISSUE #9468: Working directory not fully restored when Tabby is closed and restarted in WSL2
**Describe the problem**:

When using Tabby with WSL2, if the terminal is closed and then restored (e.g. for a Windows update), all tabs and their shell histories are restored, but after a second, the working directories are reset, and the tabs are soft-cleared (i.e. they're reset but you can still scroll up to see the past history).

**To Reproduce**:

- Open Tabby.
- Open multiple tabs and `cd` into different directories.
- Close Tabby.
- Open Tabby.
- See the tabs restored, and with correct working directories.
- See tabs reset, now with `/mnt/c/Users/$USER` working directories.

### LINKED ISSUE #11584: Windows: a file path can be used as the session CWD, causing CreateProcess to fail with error 267 (ERROR_DIRECTORY)

vjandrea:
### Problem

WSL reports its working directory to Tabby as a POSIX path (via OSC 1337 `CurrentDir=`), but `session.ts` and `terminal.service.ts` validate the cwd with `fs.existsSync()` running in the **Windows** host process, which can never resolve a POSIX path. The check silently fails and Tabby falls back to `process.env.HOME`, so **splitting a WSL tab or restoring a session lands in the Windows home dir** instead of the shell's actual directory.

Fixes #9468

### Fix

Each WSL shell now carries an `fsBase` (`\\wsl$\<Distro>` for WSL2, `<BasePath>\rootfs` for WSL1). Before validating/spawning, a POSIX cwd is rewritten against that base into a UNC path the Windows host can actually reach.

- `resolveGuestCWD(cwd, fsBase)` in `tabby-local/src/wslPath.ts` is pure and returns the cwd unchanged when it isn't a guest POSIX path or no `fsBase` is known. Ships with an `assert`-based self-check.
- `fsBase` flows `Shell` -> `SessionOptions` / `LocalProfile` -> both the split/restore path (`session.ts`) and the "open in this directory" path (`terminal.service.ts`).
- `WSLShellProvider` computes `fsBase` for the default-distro entry too; the per-distro and default paths now share one `fsBaseForDistro()` helper instead of duplicating the `Flags & 8` heuristic.

### Notes

- WSL2 vs WSL1 detection reuses the existing `Flags & 8` heuristic already in `wsl.ts`; the documented field is the `Version` DWORD. Worst case on a wrong guess is the current behaviour (fallback to HOME). Left as-is to keep the diff focused.
- Recovery tokens saved before this change have no `fsBase`, so `resolveGuestCWD` is a no-op and the fallback is the same as today. No migration needed.

### Overlap with #11592

#11592 (fixes #11584, "file path used as session CWD -> CreateProcess error 267") touches the same `fs.existsSync(cwd)` lines in `session.ts` and `terminal.service.ts`, replacing them with an `isDirectorySync()` check. The two fixes are orthogonal and composable: that PR checks whether the cwd is a real directory, this one checks whether it resolves to a real Windows path at all. Whichever lands first, the other should rebase to compose them, e.g. `isDirectorySync(resolveGuestCWD(cwd, fsBase))`. **Real merge-conflict risk on those lines; happy to coordinate.**

## AI Usage

* [ ] Fully vibe coded
* [ ] AI-designed, AI-coded, manually checked
* [X] Human-designed, AI-coded


Eugeny:
Thanks!

PR #11640, opened 2026-08-31
fix(ssh): surface SFTP upload errors

chill-DA:
## Description

SFTP uploads currently cancel the transfer when an error occurs, but the
transfer row does not expose the actual SFTP error. This can leave failed
uploads looking stuck, especially for errors such as `Permission denied`.

This change:

- stores the SFTP exception message in the transfer status before cancelling
  the upload, so the File transfers UI displays the failure reason
- ensures cancelled transfers are not considered complete just because all
  source bytes were read before a later SFTP operation failed

The second point matters for failures during finalisation, such as closing,
replacing, or renaming the temporary upload file. Previously a transfer whose
local source had already been fully read could still satisfy `isComplete()`
after being cancelled.

## Testing

- Reviewed the resulting diff against `master`
- Verified the branch contains only the intended changes in
  `tabby-core/src/api/platform.ts` and `tabby-ssh/src/session/sftp.ts`
- GitHub CI can provide the full project build/lint validation

## AI Usage

* [ ] Fully vibe coded

*<sub>This is not to block AI contributions but rather to speed up PR review
(saves time on trying to deduce the logic behind AI hallucinations).</sub>*


Eugeny:
Thanks!

PR #11649, opened 2026-09-05
Eternal Terminal support for Tabby 

qoke:
## Description

New Tabby plugin `tabby-et` that adds **Eternal Terminal** as a first-class connection type.

Why ? We all use mobile devices. Mobile devices move around. Moving around causes terminal reconnections. Terminal reconnections are a pain. ET was created to solve this pain. Tabby was created to make terminals good. Combine for more good, less pain.

Initial support for
- Eternal Terminal connection profile (host, user, ports, auth via an SSH profile)
- Quick-connect with user@host from the profile selector or the CLI
- Forward the local SSH agent to the remote session
- Connect via an ET jump host

Key design decisions:
- Native terminal scroll-back
- Use local and remote port forwarding (configured per-profile or add live)
- ET client protocol natively in TypeScript (don't exec/shell out to et)
- Use Tabby's existing SSHSession from tabby-ssh (better for security, less maintenance and more inherited features over time)
- Use tweetnacl (nacl.secretbox / nacl.secretbox.open) for crypto. Pure JS, no native build step reqd and web safe
- TERM is always xterm-256color (can iterate on this if needed)
- Hand write the protobuf codec, no protobufjs. Less deps, its only 200 lines and the et protocol isn't a moving target
- Not resuming an ET session across a Tabby restart (requires persisting the passkey and both nonce counters and a replay buffer. This is asking for security problems, I'm not implementing this)
- Not using et client configuration file - all config sits inside Tabby

## AI Usage

* [ ] Fully vibe coded
* [ ] AI-designed, AI-coded, manually checked
* [*] Human-designed, AI-coded, manually checked


Eugeny:
Thank you for the PR - it's really well done but I can't take on maintenance of an ET plugin in the mainline. 

A realistic way forward is a separate PR for the changes in `tabby-ssh` etc that you need for this to work and releasing the plugin as a third-party plugin

PR #11650, opened 2026-09-06
Fix Sogou IME commit when pressing Shift on Windows

hymsk:
## Description

Fixes dropped IME commit text on Windows when pressing <kbd>Shift</kbd> to switch Sogou Pinyin to English.

xterm.js sets its private `_keyDownSeen` fallback guard for every keydown. A pure Shift keydown can therefore suppress the following composed `input` event that carries the IME commit. This wraps xterm's `_keyDown` on Windows and restores the previous `_keyDownSeen` value for Shift-only keydowns, matching the upstream fix proposed in xtermjs/xterm.js#6054 until Tabby's xterm dependency includes it.

The check accepts both `event.key === 'Shift'` and legacy `keyCode === 16` event shapes.

### Testing

- `eslint tabby-terminal/src/frontends/xtermFrontend.ts`
- `tsc --noEmit --project tabby-terminal/tsconfig.json`
- Built the `tabby-terminal` plugin with Node.js 22 and packaged it into the official Tabby v1.0.235 Windows x64 portable binary.
- Manually verified on Windows with Sogou Pinyin: type preedit text, press Shift to switch to English, and confirm the preedit text is committed instead of being dropped.
- Verified normal PowerShell terminal startup and input in the patched build.

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/tabby-terminal/src/frontends/xtermFrontend.ts
+++ b/tabby-terminal/src/frontends/xtermFrontend.ts
@@ -116,6 +116,23 @@ export class XTermFrontend extends Frontend {
         this.flowControl = new FlowControl(this.xterm)
         this.xtermCore = this.xterm['_core']
 
+        // xterm.js#6054 does this in _keyDown itself. Keep the workaround at
+        // that boundary so Shift cannot overwrite a previous keydown's state.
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks!

PR #11651, opened 2026-09-06
Flash taskbar/dock icon when the terminal bell rings

### LINKED ISSUE #8981: Flash frame/taskbar icon on terminal bell
**Is your feature request related to a problem? Please describe.**
I am missing this feature which other terminals (Windows Terminal, putty) have - when bell comes and the terminal application does not have keyboard focus (it is minimized or another application is selected), it flashes the taskbar icon (on Windows).

When bell comes, Tabby should check config whether to flash frame and call relevant function. In Electron this can be achieved by [this api](https://www.electronjs.org/docs/latest/tutorial/windows-taskbar#flash-frame). It seems it's working on macOS as well (bouncing icon), don't know about Linux.

In addition, it would be good to change the Tabby settings to allow the user to choose more responses to the bell. See attached screenshot from Windows Terminal bellow.

There is no alternative but to not use Tabby 😄 

**Additional context**
image
Windows Terminal bell settings - *Flash window* is *Visual* in Tabby, *Flash taskbar* is this feature request

Yahiro025:
Fixes #8981

When the terminal bell rings and Tabby is not focused, flash the taskbar (Windows/Linux) or bounce the dock icon (macOS) via Electron `flashFrame`. This matches Windows Terminal’s “Flash taskbar” option.

Related but separate: #11292 (issue #11172) marks a *background tab* with `displayActivity()` (persistent blue-dot). This PR does **not** duplicate that — it only flashes the *app window* when it is unfocused.

## Description

The existing `terminal.bell` radio (`off` | `visual` | `audible`) is unchanged. A companion setting `terminal.bellFlashFrame` (default off) is added under Settings → Terminal → Sound so visual/audible and taskbar flash can be combined.

On bell, if the setting is enabled, the renderer calls `HostWindowService.flashFrame()`, which the Electron host implements over the existing window IPC path. The main process only flashes when the window is unfocused, and stops flashing when the window gains focus.

The toggle is hidden on the web platform, where `flashFrame` is a no-op.

## AI Usage

* [ ] Fully vibe coded


Eugeny:
Thanks!

Issue #11124, opened 2026-03-31
exiting an ssh session via exit command hangs the tab instead of closing it


ew4adtr:
Version: 1.0.230
Platform: Windows x64 10.0.22631
Plugins: save-output
Frontend: xterm-webgl

Running ssh to an ubuntu server - been using this setup with previous tabby versions. 
Was previously able to close a tab or tab partition by invoking exit at the shell cli. Running exit now seems to terminate the session and leave the tab or tab partition suspended. Need to close out tab containing exited session in order to get rid of it.

adude00:
I tought it was a bug too, but there is a configuration option:

- Profile and connections
- Local terminal
- When a sessione ends
- choose "Close"

Auto doesn't do anything.

Screenshot:

<img width="871" height="474" alt="Image" src="https://github.com/user-attachments/assets/0980ba59-22a3-40dd-95db-596befb9b003" />

peix187:
Same problem for me (i use Ctrl +d) with version 1.0.230.
Windows 11 25H2 26200.8037
When i close tab with the mouse the developer tools log:
```
C:\Users\MyName\Downloads\tabby-1.0.230-portable-x64\resources\builtin-plugins\tabby-core\dist\index.js:16440 [ssh-192.168.1.254-22] Destroying
C:\Users\MyName\Downloads\tabby-1.0.230-portable-x64\resources\builtin-plugins\tabby-core\dist\index.js:16440 [ssh-shell-192.168.1.254-22] Destroying
C:\Users\MyName\Downloads\tabby-1.0.230-portable-x64\resources\builtin-plugins\tabby-core\dist\index.js:16440 [ssh-192.168.1.254-22] Destroying
C:\Users\MyName\Downloads\tabby-1.0.230-portable-x64\resources\app.asar\node_modules\russh\lib\index.js:3 Unhandled Promise rejection: SendError ; Zone: <root> ; Task: Promise.then ; Value: Error: SendError Error: SendError
api.onUnhandledError @ zone.umd.js:1097
handleUnhandledRejection @ zone.umd.js:1125
_loop_2 @ zone.umd.js:1116
api.microtaskDrainDone @ zone.umd.js:1120
drainMicroTaskQueue @ zone.umd.js:641
Promise.then
nativeScheduleMicroTask @ zone.umd.js:610
scheduleMicroTask @ zone.umd.js:621
_ZoneDelegate.scheduleTask @ zone.umd.js:435
[...]

PR #11692, opened 2026-09-20
ssh: fix session sometimes not ending when remote shell exits

anseltaft:
The server is not required to send CHANNEL_EOF before CHANNEL_CLOSE and whether it does is timing-dependent (e.g. when a sudo child process delays the pty EOF). Since the switch from ssh2 to russh the shell session only listened for EOF, so when the server went straight to channel close the tab stayed connected after the remote shell exited.

Also destroy the session on channel close, matching the old ssh2 'end' event semantics which fired on both EOF and close.

## Description

Fixes issue [11124](https://github.com/Eugeny/tabby/issues/11124)

...

## AI Usage

[X] AI-designed, AI-coded, manually checked


Eugeny:
Thank you!

Eugeny:
@all-contributors add @anseltaft for code

